PR #1, opened 2021-05-17
Basic baseline

turian:
Implements a basic MelSpectrogram baseline embedding, using random projection for each embedding size.

This implements the API we have defined, but we need to make the following changes to the API document published:
* device = None and batch_size = None by default in get_audio_embedding()
* 20-dimensional we should specify is a signed int8.
* timestamps is returned as a tensor
* device argument has been removed.
* the model is copied to the audio device, not the other way around
* In pairwise distance, the embeddings should have been converted to floats already


baseline/baseline.py
```diff
@@ -0,0 +1,195 @@
...
+        x512 = x2048.matmul(self.emb512)
+        x128 = x512.matmul(self.emb128)
+        x20 = x128.matmul(self.emb20)
+        # [0, 1] range
+        x20 = self.activation(x20)
+        # Convert x20 to int8.
+        newmax = torch.iinfo(torch.int8).max
+        newmin = torch.iinfo(torch.int8).min
+        x20 = torch.tensor(
+            (x20 * (newmax - newmin) + newmin), dtype=torch.int8, device=x20.device
```
jorshi, line 55: Instead of creating a new tensor, we can cast to int8:
```
x20 = x20 * (newmax - newmin) + newmin
x20 = x20.type(torch.int8)
```

baseline/baseline.py
```diff
@@ -0,0 +1,195 @@
...
+        # Here we just pick a sensible default batch size
+        batch_size = 512
+    dataset = torch.utils.data.TensorDataset(audio)
+    loader = torch.utils.data.DataLoader(
+        dataset, batch_size=batch_size, shuffle=False, drop_last=False
+    )
+
+    if center:
+        timestamps = (
+            torch.range(0, audio.shape[1], hop_size_samples) / input_sample_rate()
```
jorshi, line 133: This is minor, so we can ignore if you want -- but I'm getting deprecation warnings for range. Suggests to update to arange. Works essentially the same, some minor differences. The epsilon on end is suggested by pytorch to handle float errors, this crops up when hop size goes into the length of audio an even number of times.
```
timestamps = torch.arange(0, audio.shape[1] + 1e-8, hop_size_samples, dtype=torch.float)
timestamps = timestamps / input_sample_rate()
```

baseline/baseline.py
maxsolomonhenry: I'm getting a warning here:
```UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).
  (x20 * (newmax - newmin) + newmin), dtype=torch.int8, device=x20.device```

File: baseline/baseline.py [hunk 3]
```diff
@@ -0,0 +1,195 @@
...
+    Args:
+            audio: n_sounds x n_samples of mono audio in the range
+            [-1, 1]. We are making the simplifying assumption that
+            for every task, all sounds will be padded/trimmed to
+            the same length. This doesn’t preclude people from
+            using the API for corpora of variable-length sounds;
+            merely we don’t implement that as a core feature. It
+            could be a wrapper function added later.
+            model: Loaded model, in PyTorch or Tensorflow 2.x. This
+        should be moved to the device the audio tensor is on.
```
maxsolomonhenry, line 102: What's the standard for the comment formatting here?

baseline/baseline.py
```diff
@@ -0,0 +1,195 @@
...
+            20].  Tensor is float32 (or signed int for 20-dim),
+            n_sounds x n_frames x dim.
+
+    """
+    assert audio.ndim == 2
+    model = model.to(audio.device)
+
+    # Implement batching of the audio
+    if batch_size is None:
+        # Here we just pick a sensible default batch size
```
maxsolomonhenry, line 124: I'm generally a fan of periods after comments, as a matter of style. (Nitpick I know. (Sorry).)

baseline/baseline.py
```diff
@@ -0,0 +1,195 @@
...
+        self.emb128 = torch.nn.Parameter(torch.rand(512, 128) / math.sqrt(512))
+        self.emb20 = torch.nn.Parameter(torch.rand(128, 20) / math.sqrt(128))
+        self.activation = torch.nn.Sigmoid()
+
+    def forward(self, x: Tensor, hop_size_samples: int, center: bool):
+        self.mel.hop_length = hop_size_samples
+        self.mel.center = center
+        self.mel.spectrogram.hop_length = hop_size_samples
+        self.mel.spectrogram.center = center
+        x = torch.log(self.mel(x) + 1e-4)
```
maxsolomonhenry, line 42: This would be a good candidate for a class variable: `eps = 1e-4` or some such thing. Adjustable, but a just a little out of reach.

baseline/baseline.py
```diff
@@ -0,0 +1,195 @@
...
+        self.mel = MelSpectrogram(
+            sample_rate=self.sample_rate, n_fft=self.n_fft, n_mels=self.n_mels
+        )
+        self.emb4096 = torch.nn.Parameter(
+            torch.rand(self.n_mels, 4096) / math.sqrt(self.n_mels)
+        )
+        self.emb2048 = torch.nn.Parameter(torch.rand(4096, 2048) / math.sqrt(4096))
+        self.emb512 = torch.nn.Parameter(torch.rand(2048, 512) / math.sqrt(2048))
+        self.emb128 = torch.nn.Parameter(torch.rand(512, 128) / math.sqrt(512))
+        self.emb20 = torch.nn.Parameter(torch.rand(128, 20) / math.sqrt(128))
```
maxsolomonhenry, line 34: And likewise here. I could see something like, (1) asserting nfft is a power of two, then rebuilding a similar structure to the layers here, but with variables instead of magic numbers. Just feels a bit more flexible.

Not tied to the suggestion. Take it or leave it.

baseline/baseline.py
```diff
@@ -0,0 +1,195 @@
...
+        )
+        return {4096: x4096, 2048: x2048, 512: x512, 128: x128, 20: x20}
+
+
+def input_sample_rate() -> int:
+    """
+    Returns:
+        One of the following values: [16000, 22050, 44100, 48000].
+            To avoid resampling on-the-fly, we will query your model
+        to find out what sample rate audio to provide it.
```
maxsolomonhenry, line 65: Again, just wondering what style convention we're using here.

baseline/baseline.py
```diff
@@ -0,0 +1,195 @@
...
+            for every task, all sounds will be padded/trimmed to
+            the same length. This doesn’t preclude people from
+            using the API for corpora of variable-length sounds;
+            merely we don’t implement that as a core feature. It
+            could be a wrapper function added later.
+            model: Loaded model, in PyTorch or Tensorflow 2.x. This
+        should be moved to the device the audio tensor is on.
+        hop_size: Number of audio samples between adjacent frames
+            batch_size: The participants are responsible for estimating
+        the batch_size that will achieve high-throughput while
```
maxsolomonhenry, line 105: See above, formatting.

baseline/baseline.py
maxsolomonhenry: +1

baseline/baseline.py
```diff
@@ -0,0 +1,195 @@
...
+            embs = model(batch[0], hop_size_samples=hop_size_samples, center=center)
+            for e in embs:
+                allembs[e].append(embs[e])
+    # Concatenate the minibatches before returning
+    # TODO: Check that returns are the right type?
+    for e in allembs:
+        allembs[e] = torch.cat(allembs[e], dim=0)
+        assert allembs[e].shape[0] == audio.shape[0]
+        assert len(timestamps) == allembs[e].shape[1]
+    return allembs, timestamps
```
maxsolomonhenry, line 157: Would prefer `embeddings` or `embedding`, likewise `all_embeddings` etc...

baseline/baseline.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+def pairwise_distance(emb1: Tensor, emb2: Tensor) -> Tensor:
+    """
+    Note that if you are calling this with the 20-dim int8,
+    you should cast them to .float() first.
```
maxsolomonhenry, line 163: Given that we're stipulating the cast to int8, maybe we should take care of this possibility ourselves.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,17 @@
...
+    rev: 0.3.9
+    hooks:
+    -   id: nbstripout
+-   repo: https://github.com/mwouts/jupytext
+    rev: v1.11.2
+    hooks:
+    -   id: jupytext
+        args: [--sync, --pipe, black]
+        additional_dependencies:
+            - black==21.5b0 # Matches hook
```
maxsolomonhenry, line 12: *Boxing glove emoji*. Bring it on.

maxsolomonhenry (commented):
A great start. Some comments about style and usability. As always, take them or leave them.

In general, I like periods after one-liner comments. I'm also pushing for more explicit variable names and points.

Wondering about having `nfft` and others as instance variables. Other general comments about making the code a bit more flexible.

baseline/baseline.py
```diff
@@ -0,0 +1,195 @@
...
+        self.mel.center = center
+        self.mel.spectrogram.hop_length = hop_size_samples
+        self.mel.spectrogram.center = center
+        x = torch.log(self.mel(x) + 1e-4)
+        x = x.swapaxes(1, 2)
+        x4096 = x.matmul(self.emb4096)
+        x2048 = x4096.matmul(self.emb2048)
+        x512 = x2048.matmul(self.emb512)
+        x128 = x512.matmul(self.emb128)
+        x20 = x128.matmul(self.emb20)
```
maxsolomonhenry, line 48: General comment... without intermediate non-linearities, I'm not sure there's much gained by doing this rather than just collapsing straight form 4096 to 20, or 4096 to 128, etc.

baseline/baseline.py
jorshi: My previous comment fixes that warning :)

baseline/baseline.py
turian: Agreed

baseline/baseline.py
turian: @maxsolomonhenry agreed that this is mathematically equivalent to just projecting from the original space to the final space. I wasn't sure which was more explicit to understand, or what would be the best practice for the dev-kit. Let me think about it.

jorshi (approved):

PR #2, opened 2021-05-21
Baseline framing

jorshi:
Builds on top of the basic mel spectrogram baseline and implements framing and batching.

### DIFF
--- a/baseline/baseline.py
+++ b/baseline/baseline.py
@@ -11,6 +11,7 @@ from typing import Any, Dict, List, Optional, Tuple
 import torch
 from torch import Tensor
 from torchaudio.transforms import MelSpectrogram
+import torch.nn.functional as F
 
 
 class RandomProjectionMelEmbedding(torch.nn.Module):
@@ -82,10 +83,55 @@ def load_model(model_file_path: str, device: str = "cpu") -> Any:
     return RandomProjectionMelEmbedding().to(device)
 
 
+def frame_audio(
+    audio: Tensor,
+    frame_size: int,
+    hop_size: int,
+    is_centered: bool
+):
+    """
+    Slice audio into equal length frames. Each adjacent frame is a hop_size number of
+    samples apart.
+
+    Args:
+        audio: input audio, expects a 2d Tensor of shape: (batch_size, num)samples)
+        frame_size: the length each resulting frame should be
+        hop_size: number of samples between frames.
+        is_centered: Pad audio by frame_size // 2 to center audio in frame
+
+    Returns:
+        A Tensor of shape (batch_size, num_frames, frame_size)
+    """
+    batch_size, num_samples = audio.shape
+
+    # Adjust the number of samples if centering to allow for a half
+    # frame size number of padded samples padded at the start.
+    start_pad = 0
+    if is_centered:
+        half_frame_size = int(frame_size // 2)
+        num_samples += half_frame_size
+        start_pad = half_frame_size
+
+    # Number of frames is the number of hops that can occur within num_samples
+    num_frames = math.ceil(num_samples / hop_size)
+
+    # Pad audio to facilitate centered frames.
+    padded_num_samples = (num_frames - 1) * hop_size + frame_size
+    end_pad = padded_num_samples - num_samples
+    audio = F.pad(audio, (start_pad, end_pad))
+
+    # Frame.
+    shape = (batch_size, num_frames, frame_size)
+    stride = (1, hop_size, 1)
+    frames = torch.as_strided(audio, shape, stride)
+
+    return frames
+
+
 def get_audio_embedding(
     audio: Tensor,
-    model: Any,
-    hop_size_samples: int,
+    model: RandomProjectionMelEmbedding,
+    hop_size: int,
     batch_size: Optional[int] = None,
     center: bool = True,
 ) -> Tuple[Dict[int, Tensor], Tensor]:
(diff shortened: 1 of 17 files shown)


baseline/baseline.py
```diff
@@ -6,54 +6,76 @@
...
-        newmin = torch.iinfo(torch.int8).min
-        x20 = torch.tensor(
-            (x20 * (newmax - newmin) + newmin), dtype=torch.int8, device=x20.device
-        )
+
+        # Scale to int8 value range and cast to int
+        int8_max = torch.iinfo(torch.int8).max
+        int8_min = torch.iinfo(torch.int8).min
+        x20 = x20 * (int8_max - int8_min) + int8_min
+        x20 = x20.type(torch.int8)
```
jorshi, line 77: I was testing this with a some different input signals and I don't think this is working quite how we expect. The ranges of the embedded signals after projection seem to be getting pretty extreme (each projection seems to emphasize that difference even more), even though it looks like the projection matrices are being normalized. So by the time we get to x20 the values are pretty only -128 or 127, we are capturing the range in that embedding very well.

baseline/baseline.py
```diff
@@ -82,79 +105,169 @@ def load_model(model_file_path: str, device: str = "cpu") -> Any:
...
+    """
+    batch_size, num_samples = audio.shape
+
+    # Adjust the number of samples if centering to allow for a half
+    # frame size number of padded samples padded at the start.
+    start_pad = 0
+    if is_centered:
+        half_frame_size = int(frame_size // 2)
+        num_samples += half_frame_size
+        start_pad = half_frame_size
```
jorshi, line 130: @maxsolomonhenry -- modified this a little bit, we were overpadding non-centered signals so I tried this approach of adjusting the num_samples by half_frame_size if centering prior to calculating the number of samples to append to the end.

Our implementation right now isn't quite in alignment with how librosa handles framing. Since we are padding the end to make sure that the entire frame from the last hop is entirely contained, we are ending up padding more at the end then I think we need to. When looking at the corresponding timestamps we end up with centers of frames that extend beyond the duration of the input. 

The way that librosa handles this is to not zero pad at all when not centering. In this case the input can be truncated a bit. When centering they pad both the start and end by half a frame and then applying framing (again no zero padding occurs during framing).

What we have here works fine, but it is a bit confusing to get a list of timestamps back that extends past the duration of the input.

turian: @jorshi Can you we use torchaudio to do this for us?

baseline/baseline.py
```diff
@@ -82,79 +105,169 @@ def load_model(model_file_path: str, device: str = "cpu") -> Any:
...
+            result = model(batch[0])
+            for size, embedding in result.items():
+                embeddings[size].append(embedding)
+
+    # Concatenate mini-batches back together and unflatten the frames back
+    # to audio batches
+    embeddings = dict(embeddings)
+    for size, embedding in embeddings.items():
+        embeddings[size] = torch.cat(embedding, dim=0)
+        embeddings[size] = embeddings[size].unflatten(0, (audio_batches, num_frames))
```
turian, line 263: It's still very hard for me to tell if this is unframing the audio correctly for each sample. I suspect it's not correct but it's hard to tell just by inspection. I think we need some sort of test :\

baseline/baseline.py
jorshi: @turian doesn't look like there are any utils for framing in torchaudio

maxsolomonhenry: @jorshi If we want a centered frame at t = 0 we need to pad up front. As for the end, if you don't want time stamps that extend beyond the signal we'll just have to truncate the output by one frame.

```
if end_pad > half_frame_size:
    # Truncate `audio` instead of padding it.

File: baseline/baseline.py [hunk 4]
```diff
@@ -82,79 +105,169 @@ def load_model(model_file_path: str, device: str = "cpu") -> Any:
...
+        center: whether of not the frames were padded to center audio, if false
+            an offset will be applied equal to frame_size // 2.
+
+    Returns:
+        Timestamps in seconds.
+    """
+
+    offset = 0
+    if not center:
+        if frame_size is None:
```
maxsolomonhenry, line 173: I don't think there's a need for this. Timestamps will be the same whether or not the frames are centered.

baseline/baseline.py
```diff
@@ -82,79 +105,169 @@ def load_model(model_file_path: str, device: str = "cpu") -> Any:
...
-    dataset = torch.utils.data.TensorDataset(audio)
+    # Split the input audio signals into frames and then flatten to create a tensor
+    # of audio frames that can be batch processed. We will unflatten back out to
+    # (audio_baches, num_frames, embedding_size) after creating embeddings.
+    frames = frame_audio(audio, model.n_fft, hop_size, center)
+    audio_batches, num_frames, frame_size = frames.shape
+    frames = frames.flatten(end_dim=1)
+
+    # We're using a DataLoader to help with batching of frames
+    dataset = torch.utils.data.TensorDataset(frames)
```
maxsolomonhenry, line 243: Dope.

baseline/baseline.py
maxsolomonhenry: @turian What do you mean by "unframing the audio"?

baseline/baseline.py
```diff
@@ -82,79 +105,169 @@ def load_model(model_file_path: str, device: str = "cpu") -> Any:
...
+    # Concatenate mini-batches back together and unflatten the frames back
+    # to audio batches
+    embeddings = dict(embeddings)
+    for size, embedding in embeddings.items():
+        embeddings[size] = torch.cat(embedding, dim=0)
+        embeddings[size] = embeddings[size].unflatten(0, (audio_batches, num_frames))
+
+    # Get timestamps in seconds.
+    timestamps = get_timestamps_for_embedding(
+        model.sample_rate, num_frames, hop_size, frame_size, center
```
maxsolomonhenry, line 267: Again, `center` is not required for this calculation.

baseline/baseline.py
maxsolomonhenry: I don't know if this is helpful, but you might just do individual projections from the largest embedding size down to each smaller one, instead of cascading projections. I suggested this a little while ago.

jorshi: Ya good idea -- i'll try this out

baseline/baseline.py
jorshi: I thought that we always calculated the timestamp for a frame as being the center of the frame -- at least for spectral stuff? Was referring to https://librosa.org/doc/latest/generated/librosa.frames_to_samples.html#librosa.frames_to_samples -- (which is used to calculate timestamps for frames).

jorshi: So if there isn't centering / zero-padding then the first timestamp will be frame_size // 2

PR #3, opened 2021-05-24
[WIP] audio preprocessing script

jorshi:
A simple command line script for batch processing audio files to convert them to the correct format and sample rate for HEAR 2021 audio tasks


jorshi:
Implemented in #8

PR #9, opened 2021-05-27
Baseline tests

khumairraj:
Added tests for the embedding and timestamps. 
For embeddings, the sliced and the batch tests are not passing. These tests check if the embeddings for audio is same when feeded in the model as part of different batches. To analyze this I have a layer-wise error analysis to see which layer the error is coming from.
For timestamps. as @jorshi suggested, there are few time stamps beyond the audio duration. This test is not passing, and we can remove this if this is expected behavior.
Please let me know in case I need to make changes, remove or add any other tests.

Also since we do not have more models yet, I have kept the test in the baseline folder. Will move it according to suggestions.


baseline/test_baseline.py
```diff
@@ -0,0 +1,306 @@
...
+        y2048 = y4096.matmul(model.emb2048)
+        assert torch.all(torch.abs(x2048[::2, ...] - y2048) < 1e-3)
+
+        x512 = x2048.matmul(model.emb512)
+        y512 = y2048.matmul(model.emb512)
+        assert torch.all(torch.abs(x512[::2, ...] - y512) < 1e-2)
+
+        x128 = x512.matmul(model.emb128)
+        y128 = y512.matmul(model.emb128)
+        assert torch.all(torch.abs(x128[::2, ...] - y128) < 1e-1)
```
khumairraj, line 250: The error rates in the assertions here are adjusted to make the tests pass. As can be seen, in the first `matmul`, the error increases to 1e-5. And subsequently, in all `matmuls`, this increases, leading to more error for lower sized embeddings, as seen from the errors which I have put to make them pass. Please let me know if this test has been structured correctly and if we need any changes here.

baseline/test_baseline.py
```diff
@@ -0,0 +1,306 @@
...
+            assert torch.all(torch.abs(output_sliced[1] - output[2]) < 1e-5)
+            assert torch.all(torch.abs(output_sliced - output[::2, ...]) < 1e-5)
+
+
+class TestLayerbyLayer:
+    def test_layers_find_error(self):
+        device = "cuda:0" if torch.cuda.is_available() else "cpu"
+        model = load_model("", device=device)
+
+        frames = torch.rand(512, model.n_fft)
```
jorshi, line 210: Same thing here, make sure on the correct device

baseline/test_baseline.py
turian: I'm not entirely sure this is something we want to test, what sort of bugs does it prevent?

tests/test_baseline.py
```diff
@@ -0,0 +1,275 @@
...
+
+    def test_embeddings_shape(self):
+        # Test the embeddings shape for centered and not centered.
+        # The embeddings size in these two cases are different by the codes
+        # logic.
+        for size, embedding in self.embeddings_not_ct.items():
+            assert embedding.shape == (64, 96000 // 256, int(size))
+
+        for size, embedding in self.embeddings_ct.items():
+            assert embedding.shape == (64, (4096 // 2 + 96000) // 256, int(size))
```
maxsolomonhenry, line 148: I think it's usual for centered framing to result in more frames.

```
# center = false

# frameNo:    1         2         3         4
#             |---------|---------|---------|---------|


# center = true

# frameNo:         1         2         3         4         5
#             [0000|----*----|----*----|----*----|----*----|0000]
                ^pad                                          ^pad
```

tests/test_baseline.py
```diff
@@ -0,0 +1,275 @@
...
+                torch.diff(self.ts_not_ct) - (self.ts_not_ct[2] - self.ts_not_ct[1])
+            )
+            < 1e-5
+        )
+
+    def test_timestamps_end(self):
+        # Test the end of the timestamp. This should technically not exceed the
+        # time of the audio. However it is hapenning in our case though?
+        assert torch.abs(self.ts_ct[-1] - 96000 / input_sample_rate()) < 1e-5
+        assert torch.abs(self.ts_not_ct[-1] - 96000 / input_sample_rate()) < 1e-5
```
maxsolomonhenry, line 188: It depends on how you want to define the centring behaviour. Either way, I made a suggestion to fix this in the comments for #2.

tests/test_baseline.py
```diff
@@ -0,0 +1,275 @@
...
+        y2048 = y4096.matmul(model.emb2048)
+        assert torch.all(torch.abs(x2048[::2, ...] - y2048) < 1e-3)
+
+        x512 = x2048.matmul(model.emb512)
+        y512 = y2048.matmul(model.emb512)
+        assert torch.all(torch.abs(x512[::2, ...] - y512) < 1e-2)
+
+        x128 = x512.matmul(model.emb128)
+        y128 = y512.matmul(model.emb128)
+        assert torch.all(torch.abs(x128[::2, ...] - y128) < 1e-1)
```
maxsolomonhenry, line 261: I suggested to @jorshi that we not cascade these, and instead project down from x onto each embedding size individually.

maxsolomonhenry:
Nice tests! It looks like the main problem is the batching. The other issues, i.e. numerical errors from cascaded projections, and centring questions, can be addressed fairly quickly I'd say.

jorshi (approved):

PR #12, opened 2021-05-28
Add stage number to Luigi workdir

turian:
This should make it easier when developing Luigi to delete the last K task outputs.

### DIFF
--- a/evaluation-tasks/util/luigi.py
+++ b/evaluation-tasks/util/luigi.py
@@ -38,7 +38,7 @@ class WorkTask(luigi.Task):
 
     def output(self):
         return luigi.LocalTarget(
-            "_workdir/%02d-done-%s" % (self.stage_number, self.name)
+            "_workdir/%02d-%s.done" % (self.stage_number, self.name)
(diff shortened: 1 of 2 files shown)


evaluation-tasks/util/luigi.py
```diff
@@ -37,14 +37,31 @@ def name(self):
         # return type(self).__name__
 
     def output(self):
-        return luigi.LocalTarget("_workdir/done-%s" % self.name)
+        return luigi.LocalTarget(
+            "_workdir/%02d-%s.done" % (self.stage_number, self.name)
```
jorshi, line 41: Nice one, really into this. This isn't specific to the PR, but how are we going to handle the _workdir when we start adding for pipelines? Should each pipeline get a separate dir inside of _workdir?

jorshi (approved):

evaluation-tasks/util/luigi.py
turian: _workdir/{config.TASKNAME}/

PR #18, opened 2021-05-28
Filter files with labels

turian:
Filter only files with labels, merges into https://github.com/neuralaudio/hear2021-eval-kit/pull/17

Closes https://github.com/neuralaudio/hear2021-eval-kit/issues/16

### DIFF
--- a/evaluation-tasks/coughvid.py
+++ b/evaluation-tasks/coughvid.py
@@ -90,10 +90,44 @@ class ExtractCorpus(WorkTask):
             pass
 
 
+class FilterLabeledMetadata(WorkTask):
+    """
+    Filter the metadata (labels) to only contain audiofiles that
+    are labeled, and save it in metadata.csv with columns:
+    filename (without extension), label
+    """
+
+    def requires(self):
+        return ExtractCorpus()
+
+    @property
+    def name(self):
+        return type(self).__name__
+
+    def run(self):
+        labeldf = pd.read_csv(
+            os.path.join(
+                self.requires().workdir, "public_dataset/metadata_compiled.csv"
+            )
+        )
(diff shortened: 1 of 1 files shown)


evaluation-tasks/coughvid.py
```diff
@@ -117,13 +151,32 @@ def name(self):
...
+
+        labeldf = pd.read_csv(
+            os.path.join(self.requires()[1].workdir, "metadata.csv"),
+            header=None,
+            names=["filename", "label"],
         )
+        filename_with_labels = list(labeldf["filename"].to_numpy())
+        assert len(filename_with_labels) == len(set(filename_with_labels))
+        filename_with_labels = set(filename_with_labels)
+
```
turian, line 168: I'm wondering if this should be split into two tasks:
1) Find filenames with labels and symlink them.
2) Subsample to number of audio files, but based upon only filenames with labels

jorshi: I'm favor of this. Would be nice to generally try to have smaller, single objective tasks

evaluation-tasks/coughvid.py
```diff
@@ -90,11 +90,45 @@ def run(self):
             pass
 
 
+class FilterLabeledMetadata(WorkTask):
+    """
+    Filter the metadata (labels) to only contain audiofiles that
+    are labeled, and save it in metadata.csv with columns:
+    filename (without extension), label
+    """
```
jorshi, line 98: This is great -- can see how this could be extended / generalized to support different types of metadata filtering. Will be super helpful.

evaluation-tasks/coughvid.py
turian: https://github.com/neuralaudio/hear2021-eval-kit/issues/24

PR #20, opened 2021-05-28
Luigi config + luigi logging level

jorshi:
Not sure we want to set the logging level to INFO quite yet -- but this makes the print out from luigi much nicer


turian:
@jorshi Wow, I haven't really been successful with Luigi logging levels, this is great.

turian (approved):

PR #33, opened 2021-05-31
Update to framing to mirror Librosa functionality

jorshi:


heareval/baseline.py
```diff
@@ -121,21 +121,27 @@ def frame_audio(audio: Tensor, frame_size: int, hop_size: int, is_centered: bool
...
+    # Number of frames that will fully fit within num_samples.
+    num_frames = 1 + (num_samples - frame_size) // hop_size
 
-    # Pad audio to facilitate centered frames.
-    padded_num_samples = (num_frames - 1) * hop_size + frame_size
-    end_pad = padded_num_samples - num_samples
-    audio = F.pad(audio, (start_pad, end_pad))
+    # Number of samples after applying framing.
+    framed_num_samples = (num_frames - 1) * hop_size + frame_size
+
```
turian, line 137: I don't get it. Why don't we just pad with frame_size // 2 at the end too?

jorshi: Need to add in comment to make this more clear -- but basically padding the end with frame_size // 2 is what we do. And then we get to framing, which truncates the padded clip to make sure that a whole number of frames can be returned. In this code instead of actually padding the input and then truncated -- we make the calculations as if we padded, and then only add the number of zeros we need.

heareval/baseline.py
```diff
@@ -121,25 +120,30 @@ def frame_audio(audio: Tensor, frame_size: int, hop_size: int, is_centered: bool
...
+    if is_centered:
+        end_pad = framed_num_samples - audio.shape[1] - start_pad
+        audio = F.pad(audio, (start_pad, end_pad))
+    else:
+        audio = audio[:, :framed_num_samples]
 
-    # Frame.
     shape = (batch_size, num_frames, frame_size)
-    stride = (1, hop_size, 1)
+    stride = (audio.stride()[0], hop_size, 1)
```
jorshi, line 146: This was the main bug, it was a bit of a mind boggle for me lol. But, audio.stride()[0] tells us how much we need to stride in order to use this in batches. For example take this tensor of shape (3,10).

```
x = torch.arange(30).view(3,10)

tensor([[ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9],
        [10, 11, 12, 13, 14, 15, 16, 17, 18, 19],
        [20, 21, 22, 23, 24, 25, 26, 27, 28, 29]])
```

And we want a frame_rate of 5 and a frame_size of 5. 

What we were doing before is this:
```
x.as_strided(size=(3, 2, 5), stride=(1, 5, 1))

tensor([[[ 0,  1,  2,  3,  4],
         [ 5,  6,  7,  8,  9]],

        [[ 1,  2,  3,  4,  5],
         [ 6,  7,  8,  9, 10]],

        [[ 2,  3,  4,  5,  6],
         [ 7,  8,  9, 10, 11]]])
```

But what we want is this:
```
x.as_strided(size=(3, 2, 5), stride=(x.stride()[0], 5, 1))

tensor([[[ 0,  1,  2,  3,  4],
         [ 5,  6,  7,  8,  9]],

        [[10, 11, 12, 13, 14],
         [15, 16, 17, 18, 19]],

        [[20, 21, 22, 23, 24],
         [25, 26, 27, 28, 29]]])

x.stride()[0] == 10
```

khumairraj: The batches were not correctly being synthesized and thus the reason for the failure of the other test case of the models as well. This is a great find.

heareval/baseline.py
```diff
@@ -35,13 +35,12 @@ def __init__(self):
...
 
         # Projection matrices.
-        self.emb4096 = torch.nn.Parameter(
-            torch.rand(self.n_mels, 4096) / math.sqrt(self.n_mels)
-        )
-        self.emb2048 = torch.nn.Parameter(torch.rand(4096, 2048) / math.sqrt(4096))
-        self.emb512 = torch.nn.Parameter(torch.rand(2048, 512) / math.sqrt(2048))
-        self.emb128 = torch.nn.Parameter(torch.rand(512, 128) / math.sqrt(512))
-        self.emb20 = torch.nn.Parameter(torch.rand(128, 20) / math.sqrt(128))
+        normalization = math.sqrt(self.n_mels)
```
turian, line 38: If we're going to do "proper" initialization, how about Xavier initialization?

turian (approved):

PR #39, opened 2021-06-03
[WIP] Baseline framerate

jorshi:
Converting the baseline API to accept frame_rate in Hz as opposed to hop_size in samples. Also, removing centering.

Some of the tests are currently broken, which we should scrutinize and fix. (Also to make sure they still make sense.)

We shouldn't merge this until the committee agrees on the switch to frame_rate from hop_size.


heareval/baseline.py
```diff
@@ -142,7 +142,7 @@ def frame_audio(
         frame_start = int(round(sample_rate * frame_number / frame_rate))
         frame_end = frame_start + frame_size
 
-        if not frame_end < num_padded_samples:
+        if not frame_end <= num_padded_samples:
```
khumairraj, line 145: The last frame can be centered at the last sample in the audio, hence the `<=` rather than the `<`.

khumairraj (commented):
`TestLayerbyLayer`. The main purpose of this test was when our batches were not being correctly made(in the last frame audio setup due to the stride) and we needed to peep into the model to confirm if the model was causing the issue. I think since the other tests(`test_embeddings_sliced' and `test_embeddings_batched` are working, we are good to remove this.

turian (approved):

tests/test_baseline.py
```diff
@@ -240,64 +182,87 @@ def test_model_sliced(self):
             assert torch.allclose(output_sliced, output[::2])
 
 
-class TestLayerbyLayer:
-    def test_layers_find_error(self):
+# class TestLayerbyLayer:
```
jorshi, line 185: Do you foresee us needing this in the future? i.e. is there a reason not to just delete this as opposed to commenting it out?

khumairraj: Yes, Removed it as the purpose of the test is already solved.

PR #48, opened 2021-06-07
Fixes to Coughvid checkpointing

turian:
This unbreaks the checkpoints, which stopped working temporarily after I added mypy tweaks

### DIFF
--- a/heareval/tasks/coughvid.py
+++ b/heareval/tasks/coughvid.py
@@ -398,7 +398,7 @@ class FinalizeCorpus(WorkTask):
     # the finalized top-level task directory
     @property
     def workdir(self):
-        return config.TASKNAME
+        return os.path.join("tasks", config.TASKNAME)
 
     def run(self):
         if os.path.exists(self.workdir):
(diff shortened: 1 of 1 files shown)


heareval/tasks/coughvid.py
```diff
@@ -389,7 +398,7 @@ def name(self):
     # the finalized top-level task directory
     @property
     def workdir(self):
-        return config.TASKNAME
+        return os.path.join("tasks", config.TASKNAME)
```
turian, line 401: I'm not sure what directory structure we want exactly, but at least all tasks should go into some directory.

heareval/tasks/coughvid.py
```diff
@@ -56,7 +56,8 @@ def run(self):
             "https://zenodo.org/record/4498364/files/public_dataset.zip",
             os.path.join(self.workdir, "corpus.zip"),
         )
-        self.output().open("w")
+        with self.output().open("w") as _:
+            pass
```
jorshi, line 60: This is an option too https://stackoverflow.com/questions/1158076/implement-touch-using-python

jorshi (approved):

heareval/tasks/coughvid.py
turian: self.output() is a LocalTarget, so you would have to grab the filename and then touch it.

PR #50, opened 2021-06-07
Eval: Cache embeddings to disk

turian:
Since many tasks might be too large to store in GPU memory (or even CPU memory), and because Wavenet-like models will be expensive at inference time, we cache all embeddings to disk.

One benefit of this approach is that since all embeddings are cached as numpy arrays, the final training code can be pytorch-only, regardless of whether the embedding model is tensorflow based.

### DIFF
--- a/heareval/task_embeddings.py
+++ b/heareval/task_embeddings.py
@@ -81,10 +81,8 @@ def get_audio_embedding_numpy(
     model: Any,
     frame_rate: float,
 ) -> Tuple[Dict[int, np.ndarray], np.ndarray]:
-    # Is there a way to avoid this .float() cast? Or do it on the
-    # numpy, not after the tensor creation?
     embedding_dict, timestamps = EMBED.get_audio_embedding(  # type: ignore
-        torch.tensor(audio_numpy, device=device).float(),
+        torch.tensor(audio_numpy, device=device),
         model=model,
         frame_rate=frame_rate,
     )
@@ -94,7 +92,7 @@ def get_audio_embedding_numpy(
     return embedding_dict, timestamps
 
 
-if __name__ == "__main__":
+def task_embeddings():
     model = EMBED.load_model(EMBEDDING_MODEL_PATH, device=device)  # type: ignore
 
     # TODO: Would be good to include the version here
@@ -124,7 +122,9 @@ if __name__ == "__main__":
                 files, labels = batch
                 audios = []
                 for f in files:
-                    x, sr = sf.read(os.path.join(task, str(embedsr), split, f))
+                    x, sr = sf.read(
+                        os.path.join(task, str(embedsr), split, f), dtype=np.float32
+                    )
                     assert sr == embedsr
                     audios.append(x)
                 audios = np.vstack(audios)
@@ -140,3 +140,7 @@ if __name__ == "__main__":
                         file_embedding_dict,
                         open(os.path.join(outdir, filename + ".pkl"), "wb"),
                     )
+
+
+if __name__ == "__main__":
+    task_embeddings()
--- a/heareval/task_embeddings.py
+++ b/heareval/task_embeddings.py
@@ -1,0 +1,132 @@
+#!/usr/bin/env python3
+"""
+Compute the embeddings for every task.
+
+TODO:
+    * Ideally, we would run this within a docker container, for
+    security.
+    * Profiling should occur here (both embedding time AFTER loading
+    to GPU, and complete wall time include disk writes).
+    * This is currently pytorch only. If we instead had a method
+    get_audio_embedding_numpy, we could instead have this work both
+    for pytorch and tensorflow.
+    https://github.com/neuralaudio/hear2021-eval-kit/issues/49
+"""
+
+import csv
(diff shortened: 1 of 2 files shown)


heareval/task_embeddings.py
```diff
@@ -0,0 +1,142 @@
...
+from typing import Any, Dict, Tuple
+
+import numpy as np
+import soundfile as sf
+import torch
+from torch.utils.data import DataLoader, Dataset
+from tqdm.auto import tqdm
+
+# This could instead be something from the participants
+EMBEDDING_PIP = "heareval.baseline"
```
jorshi, line 39: Command line arg?

heareval/task_embeddings.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def __len__(self):
+        return len(self.rows)
+
+    def __getitem__(self, idx):
+        return self.rows[idx]
+
+
+# These is a high-level wrappers on the basic API that we might
+# consider sharing among all participants.
```
jorshi, line 76: What is a high-level wrapper? The `get_audio_embedding_numpy` function?

heareval/task_embeddings.py
```diff
@@ -0,0 +1,142 @@
...
+                shuffle=True,
+            )
+            outdir = os.path.join(embeddir, task, split)
+            if not os.path.exists(outdir):
+                os.makedirs(outdir)
+            for batch in tqdm(dataloader):
+                files, labels = batch
+                audios = []
+                for f in files:
+                    x, sr = sf.read(os.path.join(task, str(embedsr), split, f))
```
jorshi, line 127: `sf.read(..., dtype=np.flaot32)` then you won't have to cast to float in the tensor above

heareval/task_embeddings.py
```diff
@@ -0,0 +1,142 @@
...
+# consider sharing among all participants.
+
+
+def get_audio_embedding_numpy(
+    audio_numpy: np.ndarray,
+    model: Any,
+    frame_rate: float,
+) -> Tuple[Dict[int, np.ndarray], np.ndarray]:
+    # Is there a way to avoid this .float() cast? Or do it on the
+    # numpy, not after the tensor creation?
```
jorshi, line 85: See comment below

heareval/task_embeddings.py
```diff
@@ -0,0 +1,142 @@
...
+    embedsr = EMBED.input_sample_rate()  # type: ignore
+
+    for task in glob.glob("tasks/*"):
+        # TODO: We should be reading the metadata that describes
+        # the frame_rate.
+        # https://github.com/neuralaudio/hear2021-eval-kit/issues/53
+        frame_rate = 10
+
+        # TODO: Include "val" ?
+        for split in ["train", "test"]:
```
jorshi, line 112: We didn't include a val in coughvid right? But we might in others? So check to see if there is a val and include it if there is?

heareval/task_embeddings.py
```diff
@@ -0,0 +1,142 @@
...
+        model=model,
+        frame_rate=frame_rate,
+    )
+    for key in embedding_dict.keys():
+        embedding_dict[key] = embedding_dict[key].detach().cpu().numpy()
+    timestamps = timestamps.detach().cpu().numpy()
+    return embedding_dict, timestamps
+
+
+if __name__ == "__main__":
```
jorshi, line 97: If we want to include this as a console script in the pip package then we should move all this into a `def main`

heareval/task_embeddings.py
turian: Yeah, that function. Could rewrite the comment. I'm also curious how we could distribute a high-level API around our API that could be used by any compatible embedder. This EMBED.xxx stuff is okay but not that amenable to wrapping cleanly.

heareval/task_embeddings.py
turian: Nice, will fix

heareval/task_embeddings.py
turian: nsynth has a val that's prespecified. I'm on the fence about using a prespecified val but I guess if it's written in stone by the dataset authors lets use it.

But it does make things less flexible if, for example, you want to do 5fold validation to pick the best hyperparameters.

I think the compromise is use val when the authors provide it, otherwise choose val during final learning (not preprocessing) however you want

heareval/task_embeddings.py
turian: good idea

heareval/task_embeddings.py
turian: sure

turian: do you prefer 'click' or 'fire' ? 'click' seems slightly more active, I've only limited experience with both

maxsolomonhenry (approved):

heareval/task_embeddings.py
turian: @jorshi I'm deferring doing this for now, since the module is global and I'm not sure whether we prefer to pass the module around as a parameter (seems cleaner?) or keep it global

PR #54, opened 2021-06-11
Add labelvocabulary.csv

turian:


heareval/tasks/README.md
```diff
@@ -20,7 +20,7 @@ For each evaluation task, the directory structure is:
 
 ## More details
 
-task.json also specifies the hop_size that we will use for the
+task.json also specifies the frame_rate that we will use for the
```
jorshi, line 23: We still need to create this json file right? We're sticking with csv files for the labels, which are non-negative ints. And then this PR introduces a new csv file that maps integer labels to the original string labels, ya?

jorshi (approved):

jorshi:
Just ran this and test.csv and train.csv are still using string labels -- these should be ints now no?

heareval/tasks/README.md
turian: We still need a file (task.json but could be something else) that specifies the task type (multiclass??) and frame_rate at the very minimum. If we don't want a README and LICENSE we could put URLs for them in here too

turian:
@jorshi Yeah I think we decided it's best to keep string labels in train and test and use labelvocabulary.csv to map them to ints.

PR #56, opened 2021-06-11
Roadmap

turian:
Here is a roadmap, which we're going to try to break down into bite-sized chunks. Please comment or message if you have any thoughts/feedback/questions


ROADMAP.md
```diff
@@ -0,0 +1,135 @@
...
+
+See [README.md](README.md) for developer getting start instructions.
+
+## Minimal complete pipeline
+
+* `heareval/baseline.py` - Simple unlearned embedding algorithm
+following the API.
+
+* `heareval/tasks/` - One time preprocessing of tasks into train +
+test WAVs and `train/test.csv`.
```
maxsolomonhenry, line 11: I might be explicit about that fact that this is to put everything into a nice common format.

ROADMAP.md
```diff
@@ -0,0 +1,135 @@
...
+
+* `heareval/predict/` - [not implemented] For each task, map from
+API-compatible embeddings to `predicted-test.csv`. (This involves
+a linear model for most task types.)
```
maxsolomonhenry, line 15: Linear model? As in just a one layer projection?

ROADMAP.md
```diff
@@ -0,0 +1,135 @@
...
+
+* `heareval/evaluate/` - Given task type, `test.csv`, and
+`predicted-test.csv` output evaluation scores.
+(Note that we can implement this now just by creating random `test.csv`
+and `predicted-test.csv` files, Christian is starting this task.)
```
maxsolomonhenry, line 20: Great!

ROADMAP.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Task Types + Prediction
+
+See [heareval/tasks/README.md](heareval/tasks/README.md) for more
+details.
+
+Ideally, `predicted-test.csv` would have identical format as
+`test.csv`. In practice, for AUC which depends upon the predicted
+score, we might need to change the format slightly.
```
maxsolomonhenry, line 29: I could see one central csv with one column that just indicates the kind of task; and the task-inappropriate data could be filled with NaN's or something. As you wish.

ROADMAP.md
```diff
@@ -0,0 +1,135 @@
...
+	    files in ranked order, we could convert it to a long
+	    set of file triples where the ordering is known.
+    	This is a bit uglier and less general but simpler to work
+    	with.
+        2) We could perform a forced ranking on all the files.
+        This makes the prediction code for ranking more gnarly,
+        but the `predict-train.csv` is clean and consistent and
+        we can just run Spearman correlation for eval.
+    * Regardless of the final output of the predictor, we will
+    probably need triplets.
```
maxsolomonhenry, line 61: I've mentioned this before, but I like the idea of a simple list of distances from each sound to one anchor file.

ROADMAP.md
```diff
@@ -0,0 +1,135 @@
...
+Here are the basic components of the heareval package.
+
+We are open to changing around the directory structure.
+
+### Baseline
+
+* `heareval/baseline.py` - A simple unlearned embedding model based
+upon Mel-spectograms and random projection. This has code for the
+entire competition API, including simple framing, etc.
+
```
maxsolomonhenry, line 74: TODO, confirm that this setup will work (up to and including eval) for a known "quality" embedding; this will establish wether perfucntory learning at the end is a good appraisal of embedding quality.

ROADMAP.md
turian: @maxsolomonhenry sure why not, we can easily add a hidden layer later. Main point is to try to get something e2e done

ROADMAP.md
turian: @maxsolomonhenry it gets a little more tricky than this in the multi-label scenario. I was chatting with @csteinmetz1 about this a bit.

Because for the predicted-test.csv, we could conceivably have a unbounded float score (or [0, 1] probability) for all---say---128 midi notes.

ROADMAP.md
```diff
@@ -0,0 +1,135 @@
...
+### Learning Based Tasks
+
+Entire audio, multiclass:
+    * Concat all embeddings for the audio, make a multiclass prediction.
+
+Entire audio, multilabel:
+    * Concat all embeddings for the audio, make a multilabel prediction.
+
+Frame-based multilabel (transcription + sound event detection):
+    * For each frame, make a multilabel prediction.
```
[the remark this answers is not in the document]
turian: I think all three can be clustered actually :)

ROADMAP.md
turian: Which anchor file? The middle one? Should we use prior knowledge and cheat a bit when doing the ranking?

We could also do the ranking versus ALL the files, and average the ranks.

csteinmetz1 (approved):

ROADMAP.md
maxsolomonhenry: The "lowest" or "highest" incidence file, I guess? Assume one-dimension of variation (which I guess is implicit in a ranking task anyway, no?)

ROADMAP.md
turian: https://github.com/neuralaudio/hear2021-eval-kit/issues/67

ROADMAP.md
```diff
@@ -24,27 +24,34 @@ and `predicted-test.csv` files, Christian is starting this task.)
...
+    * The output format of `predicted-test.csv` is:
+    ```
+    filename,timestamp,label 0 probability,label 1 probability,...
+    ```
 
 ### Unlearned tasks
 
 JND (audio1 vs audio2 has a perceptible difference):
     * np.abs(embedding1 - embedding2), where embedding1 is the
     concatenation of all audio1 frames, same for embedding2.
```
maxsolomonhenry, line 49: We're confirmed on the concatenated embeddings idea?

PR #58, opened 2021-06-12
Google speech command

jorshi:
Starting the pre-processing pipeline for Google Speech Command evaluation dataset


heareval/tasks/config/speech_commands.py
```diff
@@ -0,0 +1,36 @@
+"""
+Configuration for the google speech commands task
+"""
+
+# TODO: move some of these to a global config and import that here
+# See: https://github.com/neuralaudio/hear2021-eval-kit/issues/10
+
+TASKNAME = "speech_commands-v0.0.2"
+
+# Number of CPU workers for Luigi jobs
```
turian, line 10: This should be a generic luigi config I think

heareval/tasks/config/speech_commands.py
```diff
@@ -0,0 +1,36 @@
...
+NUM_WORKERS = 4
+
+# If you only use one sample rate, you should have an array with
+# one sample rate in it.
+# However, if you are evaluating multiple embeddings, you might
+# want them all.
```
turian, line 16: Same

heareval/tasks/config/speech_commands.py
```diff
@@ -0,0 +1,36 @@
...
+SAMPLE_RATES = [48000, 44100, 22050, 16000]
+
+# TODO: Pick the 75th percentile length?
+SAMPLE_LENGTH_SECONDS = 8.0
+
+# TODO: Do we want to call this FRAME_RATE or HOP_SIZE
```
turian, line 22: Let's standardize on frame_rate everywhere?

heareval/tasks/speech_commands.py
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env python3
```
turian, line 1: Rename to Google Speech Commands? Or what is the official name of the dataset?

heareval/tasks/util/luigi.py
```diff
@@ -6,6 +6,8 @@
 import os
```
turian, line 6: Nice abstractions

heareval/tasks/util/luigi.py
```diff
@@ -63,6 +77,55 @@ def stage_number(self) -> int:
...
 
+class DownloadCorpus(WorkTask):
+    """
+    Task for downloading a dataset to a specific filename
+    """
+
+    # URL to download dataset from
+    url = luigi.Parameter()
+    outfile = luigi.Parameter()
+
```
turian, line 88: Okay so you have to be really really careful here because this is a gotcha this is hard for us to detect automatically through code.

The checkpoint name is based upon self.name. If there are parameters and they are not used in self.name, then the checkpoints won't track multiple calls to this task. In this case, the task name is made unique by slugifying the in url.

heareval/tasks/util/luigi.py
```diff
@@ -63,6 +77,55 @@ def stage_number(self) -> int:
...
+        return 0
+
+
+class ExtractCorpus(WorkTask):
+
+    infile = luigi.Parameter()
+
+    @property
+    def name(self):
+        return type(self).__name__
```
turian, line 109: Same as above. I'm curious if there's a way to automagically grab all parameters and put them in the name by default, unless you really specifically want to override that.

heareval/tasks/util/luigi.py
```diff
@@ -63,6 +77,55 @@ def stage_number(self) -> int:
...
+
+        filename, extension = os.path.split(self.infile)
+
+        # TODO -- run correct extraction given archive type
+        print(extension)
+
+        # Location of zip file to extract. Figure this out before changing
+        # the working directory.
+        corpus_zip = os.path.realpath(
+            os.path.join(self.requires().workdir, "corpus.zip")
```
turian, line 121: I'm confused, how did we get from self.infile to requires().workdir/corpus.zip?

Also, I think we should check the file extension and make this called ExtractZipCorpus

heareval/tasks/util/luigi.py
turian: I mean maybe if we call it uid_including_parameters it might also be clearer?

turian: They have some notion of insignificant parameters, https://luigi.readthedocs.io/en/stable/parameters.html#insignificant-parameters but my understanding is that those are only applicable when running things from scratch, to determine they should not be part of the task signature. They won't fix the issue if you are re-running an incomplete pipeline.

heareval/tasks/util/luigi.py
khumairraj: We have already linked the `DownloadCorpus` to `ExtractCorpus`. I don't think we need an input to the `ExtractCorpus` since the file can be fetched from the `DownloadCorpus.workdir`.

heareval/tasks/util/luigi.py
turian: @khumairraj this is an important gotcha and curious also your thoughts here on how to protect against this

heareval/tasks/util/luigi.py
```diff
@@ -63,6 +77,55 @@ def stage_number(self) -> int:
...
+
+    @property
+    def stage_number(self) -> int:
+        return 0
+
+
+class ExtractCorpus(WorkTask):
+
+    infile = luigi.Parameter()
+
```
turian, line 106: Where is the self.requires for this method?

heareval/tasks/util/luigi.py
turian: @khumairraj where do you see "We have already linked the DownloadCorpus to ExtractCorpus"? I don't see a self.requires for this class.

It seems to me that the correct behavior is self.requires() based upon a prerequisite DownloadCorpus.

In this case, the issue of checkpoint name is even more dangerous.

Let's say we have DownloadCorpus(file1) and DownloadCorpus(file2). They write checkpoint: 00-DownloadCorpus-file1 and 00-DownloadCorpus-file2 and if those files exist then we know that we have downloaded those file. And we have two ExtractCorpus tasks (with NO parameters but a separate DownloadCorpus requires for each). Using our existing pattern, the checkpoint will be called 01-ExtractCorpus. i.e. both tasks will have the same checkpoint name!

This is definitely making me quite worried. I'm almost wondering if we should switch to joblib which handles dependency management more cleanly I think.

@jorshi @khumairraj ^^

heareval/tasks/util/luigi.py
khumairraj: Currently, it is in the speech_commands.py. `extract.require = lambda...`. I am working on this method though. Have some changes. will push.

heareval/tasks/util/luigi.py
khumairraj: The checkpoints are saved inside subdirectories of each task. For example `speech_commands-v0.0.2/00-DownloadCorpus.done`. The checkpoint for another task will be in its own subdirectory. WIll its still be a problem?

heareval/tasks/util/luigi.py
khumairraj: We are currently saving the checkpoints inside each task subdirectory. This will prevent this problem to happen. What do you think?

heareval/tasks/util/luigi.py
turian: Let's say that speech commands has two DownloadCorpus tasks. You run the pipeline and cancel it halfway through and only one DownloadCorpus command has finished. What happens then...

heareval/tasks/speech_commands.py
```diff
@@ -2,40 +2,46 @@
...
 import heareval.tasks.util.luigi as luigi_util
-from heareval.tasks.util.luigi import WorkTask
+from heareval.tasks.util.luigi import DownloadCorpus, WorkTask, ExtractZipFile
 
 
 # Set the task name for all WorkTasks
 WorkTask.task_name = config.TASKNAME
 
 
+class DownloadExtractCorpus(ExtractZipFile):
```
khumairraj, line 22: This adds the requirement of download the extract zip file. Was confused about how to get the output location from the download to the extract. Though of this as a way. Please recommend if anything else is suitable.

heareval/tasks/util/luigi.py
```diff
@@ -42,7 +45,8 @@ def name(self):
         # return type(self).__name__
 
     def output(self):
-        f = os.path.join(self.task_subdir, f"{self.stage_number:02d}-{self.name}.done")
+        #Replace the name with task_id as it is unique at task and parameter level
+        f = os.path.join(self.task_subdir, f"{self.stage_number:02d}-{self.task_id}.done")
```
khumairraj, line 49: Added task_id in place of the name. The task id uses the parameters from the task and hashes them together to get a unique string for task parameter combinations. This is to solve the problem of the reusability of the Luigi task.

heareval/tasks/util/luigi.py
```diff
@@ -100,32 +104,33 @@ def stage_number(self) -> int:
...
-
         # Location of zip file to extract. Figure this out before changing
         # the working directory.
         corpus_zip = os.path.realpath(
-            os.path.join(self.requires().workdir, "corpus.zip")
+            os.path.join(self.requires().workdir, self.infile)
         )
-        subprocess.check_output(["unzip", "-o", corpus_zip, "-d", self.workdir])
-
+        shutil.unpack_archive(corpus_zip, self.workdir)
```
khumairraj, line 130: Using `shutil.unpack_archive`. I checked it can automatically infer the type of the zip file and unzip.

heareval/tasks/util/luigi.py
jorshi: Nice one! I like this solution. might also be nice to have the name in there too, if not just to help with debugging -- as long as it doesn't make the file name too long.

Did you mean to remove the luigi.LocalTarget return type?

heareval/tasks/util/luigi.py
```diff
@@ -63,6 +81,56 @@ def stage_number(self) -> int:
...
+    infile = luigi.Parameter()
+    
+    @property
+    def name(self):
+        return type(self).__name__
+
+    def requires(self):
+        #Some Task should be returned having a working directory from 
+        #which the infile will be picked.
+        raise NotImplementedError
```
jorshi, line 118: I don't love enforcing that this class needs to be overridden. Unless I'm missing something about how Luigi works under the hood (i.e. the requires must be set prior to instantiation for the task_id or something) -- I'm a bit more in favour of being about to just use these out of the box and set the requires using a lambda. Seems more efficient to me.

heareval/tasks/util/luigi.py
```diff
@@ -63,6 +81,56 @@ def stage_number(self) -> int:
...
+        download_file(self.url, os.path.join(self.workdir, self.outfile))
+        with self.output().open("w") as _:
+            pass
+
+    @property
+    def stage_number(self) -> int:
+        return 0
+
+
+class ExtractZipFile(WorkTask):
```
jorshi, line 107: If unpack archive works for many different archive types lets update this name to something more generic like `ExtractArchive` and update variable names below that refer to a zip

heareval/tasks/speech_commands.py
khumairraj: As discussed issue with pickling lambda function. Possible ways is the list constructor options or passing the task as a `prev_task = luigi.TaskParameter()` and using it in requires like `self.prev_task`

heareval/tasks/util/luigi.py
khumairraj: The task_id has the name attached to it. It goes like `name_params_hash`.

heareval/tasks/util/luigi.py
```diff
@@ -107,23 +107,21 @@ def stage_number(self) -> int:
 class ExtractArchive(WorkTask):
     
     infile = luigi.Parameter()
+    #The previous task will be passed in as a parameter.
+    #This must have workdir attribute and the zip file to extract should be stored inside it.
+    prev_task = luigi.TaskParameter()
```
khumairraj, line 112: Added a `luigi.TaskParameter` as input to take the previous task. This might solve the problem of reusing the tasks from the utils.

heareval/tasks/speech_commands.py
```diff
@@ -0,0 +1,39 @@
...
+# Set the task name for all WorkTasks
+WorkTask.task_name = config.TASKNAME
+
+
+def main():
+    print("max_files_per_corpus = %d" % config.MAX_FILES_PER_CORPUS)
+    luigi_util.ensure_dir("_workdir")
+    
+    download_corpus = DownloadCorpus(url = config.DOWNLOAD_URL, outfile = 'corpus.tar.gz')
+    extract_download_corpus = ExtractArchive(infile = 'corpus.tar.gz', prev_task = download_corpus)
```
jorshi, line 27: @khumairraj do have the pre-commit hooks installed? Looks like black hasn't been run on this code

heareval/tasks/util/luigi.py
turian: Nice

heareval/tasks/util/luigi.py
turian: I really don't think we should do this. We can just return DownloadCorpus(url) in self.requires()

khumairraj:
I have added the tasks for the Google Speech command and made the parts as general as possible. One thing to assist in this is the `ConfigureProcessMetaData` which will help configure the process metadata which will include slug, test train split, hash, and other data-specific tasks. Other changes follow with this flow.  Also merged the trim and convert to mono in one part with FFmpeg. Other changes have been added as comments at important parts. This is not the final version and any comments and suggestions would be great help. We can go back to anything if something doesn't make sense. Thanks

heareval/tasks/speech_commands.py
khumairraj: Done

heareval/tasks/util/luigi.py
khumairraj: Changed in the new commit.

jorshi (commented):
@khumairraj -- this is looking awesome! One thing I am thinkin about is that the way the data is being labelled is a bit different than how we are are going to want it ultimately. We want to follow what they do in the [tf dataset](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/audio/speech_commands.py) -- there are 10 recognized commands and then the rest are labelled as unknown plus there are silent clips with background noise. I'm wondering if we should use the tensorflow dataset loader here so we get the correct train/val/test split that is what they used originally. Maybe there is a way to include that in the beginning stages of the pipeline to download and create the splits? Then we can subsample that (preferably in a stratified way) and to the resampling? What do you think @turian ?

heareval/tasks/speech_commands.py
```diff
@@ -2,38 +2,183 @@
...
+        process_metadata["filename_hash"] = (
+            process_metadata["slug"]
+            .apply(lambda relpath: re.sub(r"-nohash-.*$", "", relpath))
+            .apply(filename_to_int_hash)
+        )
+
+        # Get the data partition either from the input file or by doing the hash
+        process_metadata["partition"] = process_metadata["filename_hash"].apply(
+            partial(which_set, validation_percentage=0.0, testing_percentage=10.0)
+        )
```
khumairraj, line 88: The `filename_hash` and the `partition` are made as part of the `Configure Process Metadata`. The reason to do it here is that, a new dataset might have another test train split strategy and that we can capture that in the `Configure Process Metadata`. I though of doing the sampling of the configuration data here as well so that we can stratify it based on the label. This was we will have the partition name, label for the all the subsampled files and in the next step we can go ahead with the next steps.

heareval/tasks/speech_commands.py
```diff
@@ -2,38 +2,183 @@
...
+        # In this case we take the slug and remove the -nohash- as described
+        # This nohash removal allows for the speech of similar person to be in the
+        # same dataset. Such type of data specific requirements might be there.
+        # in the readme of google speech commands. we want to keep similar people
+        # in the same group - test or train or val
+        process_metadata["filename_hash"] = (
+            process_metadata["slug"]
+            .apply(lambda relpath: re.sub(r"-nohash-.*$", "", relpath))
+            .apply(filename_to_int_hash)
+        )
```
khumairraj, line 83: Google speech command also works by this method to sample the train test groups. They remove the no-hash part and then hash the file so that same speaker person will have a same hash and so that they fall in the same partition(`train` `test` or `val`). Its just that here i am applying the hash on the slug which is `label+filename minus the hash part`(reason being that similar persons have participated in the data generation for each label, but this is a choice). Can change it to the way done in the Google speech command which is something like `filename minus the hash part`. Is this strategy of doing everything in the `Configure process metadata` correct? The main goal here being that we can configure these `process level metadata` which are dependent on the `dataset` we are making the `pipeline` for and then we can plug in generic tasks which will probably take other task folders and this `process metadata` as input to produce subsequent output. @jorshi We can change the hash part for doing the split similar to the google speech command to get similar test train split and for the stratified subsampling, we can subsample this config file here so that the `SubSampleCorpus` can use this subsampled `process metadata`. This is how we can think about it in this setup. Please let me know if this is a good way?

heareval/tasks/speech_commands.py
```diff
@@ -2,38 +2,183 @@
...
+
+    def requires(self):
+        return ExtractArchive(infile="corpus.tar.gz")
+
+    def run(self):
+        # Get relative path of the audio files
+        # This file can also be built with the metadata file for the dataset
+        # in case the metadata is provided
+        process_metadata = pd.DataFrame(
+            glob(os.path.join(self.requires().workdir, "[!_]*/*.wav")),
```
khumairraj, line 56: As part of the `ConfigureProcessMetaData`, the metadata file is made which pulls up all the files which will be used for training and testing. Since this is data specific as different data can have different folder structures and different ways to get these files, we get the relative path of those files and add it to the metadata so that going ahead these files can be treated as just training points. @jorshi To make background data to not be part of the training data, I have added the regex `[!_]*/*.wav` so that the `_background` folder is not used while building the training points list. I have made an assumption that the files in this will not be part of the training or testing data. But as I understand if we need to make another category as `background` by doing some processing, we can do it here as well. This is kind of the design of the flow that makes all these decisions in this class only leaving the subsequent tasks to do the heavy lifting of executing the process with the data-specific decisions made by this class.

heareval/tasks/config/speech_commands.py
jorshi: Added a global dataset_configy.py file

heareval/tasks/config/speech_commands.py
```diff
@@ -18,21 +21,9 @@
...
-# Most of the tasks iterate over every audio file present,
-# except for the one that downsamples the corpus.
-# (This is why we should have one working directory per task)
-MAX_FRAMES_PER_CORPUS = 20 * 3600
-
-MAX_FILES_PER_CORPUS = int(MAX_FRAMES_PER_CORPUS / FRAME_RATE / SAMPLE_LENGTH_SECONDS)
+# Number files to include -- this is the full dataset
+MAX_TRAIN_FILES = 85511
+MAX_VAL_FILES = 10102
+MAX_TEST_FILES = 4890
```
turian, line 18: Can we just use None and include these numbers as a comment?

heareval/tasks/speech_commands.py
```diff
@@ -21,111 +23,255 @@
...
+                url=config.TEST_DOWNLOAD_URL, outfile="test-corpus.tar.gz"
+            )
         }
 
 
+class GenerateTrainDataset(WorkTask):
+    """
+    Silence / background samples in the train / validation sets need to be
+    created by slicing up longer background samples into 1sec slices
+    """
```
turian, line 63: Wait what? Why? Is there a cite for this?

If not, can we avoid this or skip this?

turian (approved):
Generally quite nice. See a few quick comments below.

I am curious how much of the speech commands stuff can be thought of as generic Luigi stuff

heareval/tasks/speech_commands.py
jorshi: Ya this is the intention for the dataset. It is shipped with long 1min+ background noise audio for the training and val sets that are supposed to be cut up in 1s chunks. This is the exact method that is used in the tf dataset: https://github.com/tensorflow/datasets/blob/79d56e662a15cd11e1fb3b679e0f978c8041566f/tensorflow_datasets/audio/speech_commands.py#L142

The test set ships with a set of predefined 1s background samples.

turian: @jorshi okay add this information to the comment?

jorshi: Will add this as a comment

jorshi:
Merged -- will work on genericizing components of this in a new PR

PR #64, opened 2021-06-15
[WIP] Training of baseline

turian:
There are a handful of things that need to be cleaned up here, which I'll add comments on.

Another thing that is a little weird is our embeddings don't have any activation function, so applying a second linear layer means it's really just a linear projection from mel to labels. We might actually consider putting a squashing function on the baseline embeddings or see what the last layer of openl3 is. (Adding a squashing function over the embeddings helps a lot on our naive baseline, but most models will probably already be squashed or normalized, so I don't leave it on by default.)

- [ ] Add to tests


heareval/train_baseline.py
```diff
@@ -0,0 +1,164 @@
...
+        return optimizer
+
+
+# %load_ext tensorboard
+# %tensorboard --logdir lightning_logs
+
+# +
+# init model
+model = MulticlassEmbedding()
+
```
turian, line 159: We should also add validation, early stopping, and test output

heareval/tasks/coughvid.py
```diff
@@ -366,8 +366,24 @@ def run(self):
         labelcsv = csv.writer(
             open(os.path.join(self.workdir, "labelvocabulary.csv"), "wt")
```
turian, line 367: This could be a separate PR. It maps the labels in train and test to the vocabulary numbers for the final task output

heareval/train_baseline.py
```diff
@@ -0,0 +1,164 @@
...
+from torch.utils.data import DataLoader, random_split
+
+import heareval.baseline
+
+from csvdataset import CSVDataset
+from heareval.baseline import RandomProjectionMelEmbedding
+
+
+TASK = "coughvid-v2.0.0"
+
```
turian, line 24: This stuff should all be pulled from task metadata

heareval/train_baseline.py
```diff
@@ -0,0 +1,164 @@
...
+FRAME_RATE = 8
+SAMPLE_LENGTH_SECONDS = 8.0
+N_LABELS = 3
+
+# TODO: Support for multiple GPUs?
+device = "gpu" if torch.cuda.is_available() else "cpu"
+
+# +
+VALIDATION = 0.2
+
```
turian, line 36: This should be grabbed from the model

heareval/train_baseline.py
```diff
@@ -0,0 +1,164 @@
...
+import numpy as np
+import pytorch_lightning as pl
+import soundfile as sf
+import torch
+from torch.utils.data import DataLoader, random_split
+
+import heareval.baseline
+
+from csvdataset import CSVDataset
+from heareval.baseline import RandomProjectionMelEmbedding
```
turian, line 20: We're using heareval.baseline here, but this should be generic and work with any model that follows our API.

heareval/train_baseline.py
```diff
@@ -0,0 +1,164 @@
...
+# TODO: Support for multiple GPUs?
+device = "gpu" if torch.cuda.is_available() else "cpu"
+
+# +
+VALIDATION = 0.2
+
+AUDIO_BATCH_SIZE = 64
+FRAME_BATCH_SIZE = 64
+EMBEDDING_SIZE = 4096
+
```
turian, line 40: We should be able to run this both in fine-tuning mode, and non-fined-tuned prediction mode.

heareval/train_baseline.py
```diff
@@ -0,0 +1,164 @@
...
+# -
+
+
+class MulticlassEmbedding(pl.LightningModule):
+    SR = heareval.baseline.input_sample_rate()
+
+    def __init__(self):
+        super().__init__()
+        self.embedding = RandomProjectionMelEmbedding()
+        # Add one because the start and end are framed
```
turian, line 79: We might also want to add a test that the models all return the expected number of frames, as computed here.

heareval/train_baseline.py
```diff
@@ -0,0 +1,164 @@
...
+        audio = []
+        for f in filenames:
+            x, sr = sf.read(
+                os.path.join(os.path.join("../tasks", TASK, str(self.SR), split, f)),
+                dtype=np.float32,
+            )
+            assert sr == self.SR
+            audio.append(x)
+        audio = torch.tensor(np.vstack(audio), device=device)
+
```
turian, line 97: This is pretty gross and suggests maybe we should change the API more.

I am not sure if we call heareval.baseline.get_audio_embedding() if gradients for the model will be accumulated

heareval/baseline.py
```diff
@@ -153,6 +153,7 @@ def get_audio_embedding(
     model: RandomProjectionMelEmbedding,
     frame_rate: float,
     batch_size: Optional[int] = 512,
+    disable_gradients: bool = True,
```
turian, line 156: Possible API addition?

PR #66, opened 2021-06-16
Baseline update

jorshi:
Updates to the current baseline model to and follow recent API updates:

- Model returns a single embedding now as opposed to a dictionary of embeddings
- `load_model()` returns a tuple of the loaded Model as well as a dictionary containing information including the sample_rate, embedding_size, and model version.
- prepping `load_model()` to also support loading learned model that @Luckygyana is working on
- `input_sample_rate()` has been removed since the sample_rate is now being returned in load_model()

### DIFF
--- a/heareval/baseline.py
+++ b/heareval/baseline.py
@@ -36,14 +36,9 @@ class RandomProjectionMelEmbedding(torch.nn.Module):
 
         # Projection matrices.
         normalization = math.sqrt(self.n_mels)
-        self.emb4096 = torch.nn.Parameter(torch.rand(self.n_mels, 4096) / normalization)
-        self.emb2048 = torch.nn.Parameter(torch.rand(self.n_mels, 2048) / normalization)
-        self.emb512 = torch.nn.Parameter(torch.rand(self.n_mels, 512) / normalization)
-        self.emb128 = torch.nn.Parameter(torch.rand(self.n_mels, 128) / normalization)
-        self.emb20 = torch.nn.Parameter(torch.rand(self.n_mels, 20) / normalization)
-
-        # An activation to squash the 20D embedding to a [0, 1] range.
-        self.activation = torch.nn.Sigmoid()
+        self.projection = torch.nn.Parameter(
+            torch.rand(self.n_mels, 4096) / normalization
+        )
 
     def forward(self, x: Tensor):
         # Compute the real-valued Fourier transform on windowed input signal.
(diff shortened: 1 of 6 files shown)


heareval/baseline.py
```diff
@@ -58,50 +53,63 @@ def forward(self, x: Tensor):
...
     In this baseline, we don't load anything from disk.
 
     Args:
-        model_file_path: Load model checkpoint from this file path.
-            device: For inference on machines with multiple GPUs,
+        model_file_path: Load model checkpoint from this file path. For this baseline
+            there is a basic version that contains no learning. To use the basic model
+            pass in 'basic' as the path. To load a random-init version of the learned
+            baseline pass in 'learned'. Passing in the location of a saved learned
+            baseline model will attempt to load that from disk.
```
turian, line 73: Hmm this still seems kinda gross to me. Because you can do learning on the "basic" model and train the random projection to become a learned projection. This also kinda sets a bad example for how the API should be used. Should be reconsider and put everything in the class?

turian: Then we can subclass from FrameBasedEmbeddingModel for framing?

heareval/task_embeddings.py
```diff
@@ -72,33 +71,30 @@ def __getitem__(self, idx):
...
 
 def task_embeddings():
-    model = EMBED.load_model(EMBEDDING_MODEL_PATH, device=device)  # type: ignore
+    model, meta = EMBED.load_model(EMBEDDING_MODEL_PATH, device=device)  # type: ignore
 
     # TODO: Would be good to include the version here
     # https://github.com/neuralaudio/hear2021-eval-kit/issues/37
     embeddir = os.path.join("embeddings", EMBED.__name__)  # type: ignore
-    embedsr = EMBED.input_sample_rate()  # type: ignore
+    embed_sr = meta["sample_rate"]
```
turian, line 97: Here's another example where doing model.sample_rate seems to make more sense than having a model and a meta

turian:
I appreciate the dict simplification but I am still thinking some of the API is unappealing. Let's talk?

heareval/task_embeddings.py
turian: On the other hand, I'm wondering if putting it into the class poses some unnecessary restriction on a particular exotic model that I can't imagine right now. I don't think so though

heareval/baseline.py
turian: Alternately, we could just stick to the one module per model idea. And have heareval.baseline.linear and heareval.baseline.convnet

heareval/model/baseline.py
```diff
@@ -15,9 +14,16 @@
 
 
 class RandomProjectionMelEmbedding(torch.nn.Module):
+    # sample rate, embedding size, and version number are
+    # required model attributes for the HEAR API
+    # TODO how do we want to handle version?
+    sample_rate = 44100
+    embedding_size = 4096
+    version = "0.0.1"
```
turian, line 22: Let's take out the version number and make an issue for it. The version number I think will be associated with code AND a .pt file, not just code

turian (approved):

PR #68, opened 2021-06-19
hop_size is back

turian:


heareval/model/baseline.py
```diff
@@ -133,7 +132,7 @@ def frame_audio(
 def get_audio_embedding(
     audio: Tensor,
     model: torch.nn.Module,
-    frame_rate: float,
+    hop_size: float,
```
jorshi, line 135: Since discussions to allow hop_size to be defined by participants, let's just assign a default value for this? Say 25ms?

jorshi (commented):
Overall looks good -- I know this is going to change based on API updates. But maybe we merge this since the hop_size will be helpful here, and then add all baseline updates in a new PR?

heareval/model/baseline.py
jorshi: Also do we want this value to be in ms as opposed to seconds?

jorshi:
@turian -- should we pull out all the models from this repo since we have the separate repo for the baseline now?

turian:
@jorshi "should we pull out all the models from this repo since we have the separate repo for the baseline now?" yeah

jorshi:
Closing this since the baseline has been moved to a new repo

PR #70, opened 2021-06-23
Speech command eval

jorshi:
Top-1 error evaluation on test ground truth labels and random frame-wise predictions.


turian (approved):
Right now, since we're switching to framewise vs scenewise embeddings, I don't think we can accept this unless it's reworked. @jorshi thoughts?

turian:
OOf didn't mean to accept this :\ not sure how to undo that

jorshi:
@turian -- ya this needs to be updated to handle scene predictions

jorshi:
Implemented in #143

PR #72, opened 2021-07-06
Google Speech Commands - TFDS

jorshi:
Do we want to use the tensorflow dataset for google speech commands? One of the benefits of this approach is that it sets up the dataset with the labels: `['down', 'go', 'left', 'no', 'off', 'on', 'right', 'stop', 'up', 'yes', '_silence_', '_unknown_']` which is how evaluation was proposed in https://arxiv.org/abs/1804.03209


jorshi:
replaced by #73

PR #73, opened 2021-07-16
Speech Commands V2

jorshi:
Update to speech commands to match the TFDS. Tried using the TFDS builder for this, but it was getting really hacky. So this implements the same approach with the 10 known commands and then labeling everything else as unknown, plus generates clips of silence from the background noise for the training and validation sets.


heareval/tasks/util/luigi.py
```diff
@@ -7,14 +7,15 @@
 import shutil
 import zipfile  # Required for shutil to work for tar.gz
```
jorshi, line 8: this is causing a flake error -- this just needs to be imported for shutil?

heareval/tasks/util/audio.py
```diff
@@ -47,21 +49,23 @@ def convert_to_mono_wav(in_file: str, out_file: str):
 
 
 def resample_wav(in_file: str, out_file: str, out_sr: int):
-    # TODO: Don't do anything if we are already the right sample rate.
```
jorshi, line 50: I don't think this should be responsible for handling that. But this is a good idea to implement in the resample task perhaps? Would involve knowing the sample rate beforehand.

heareval/tasks/util/audio.py
```diff
@@ -6,24 +6,26 @@
...
     ret = subprocess.call(
         [
             "ffmpeg",
             "-y",
             "-i",
-            in_file,
+            str(in_file),
             "-filter_complex",
-            f"apad=whole_dur={str(min_dur)}",
+            f"apad=whole_dur={duration},atrim=end={duration}",
```
jorshi, line 20: Need to trim to the correct length too.

heareval/tasks/speech_commands.py
```diff
@@ -82,50 +184,182 @@ def run(self):
...
 
 
-class MonoWavTrimCorpus(MonoWavTrimCorpus):
+class SubsamplePartitions(WorkTask):
+    """
+    Aggregates subsampling of all the partitions into a single task as dependencies.
+    All the subsampled files are stored in the requires workdir, so we just link to
+    that since there aren't any real outputs associated with this task.
+    This is a bit of a hack -- but it allows us to avoid rewriting
+    the Subsample task as well as take advantage of Luigi concurrency.
```
jorshi, line 327: This felt a bit hacky with the symlink on the entire workdir -- but maybe it is okay

heareval/tasks/config/speech_commands.py
```diff
@@ -18,21 +21,16 @@
...
-# except for the one that downsamples the corpus.
-# (This is why we should have one working directory per task)
-MAX_FRAMES_PER_CORPUS = 20 * 3600
+MAX_HOURS = 10.0
 
-MAX_FILES_PER_CORPUS = int(MAX_FRAMES_PER_CORPUS / FRAME_RATE / SAMPLE_LENGTH_SECONDS)
+MAX_FILES_PER_CORPUS = int(MAX_HOURS * 60.0 * 60.0 / SAMPLE_LENGTH_SECONDS)
+
+# Number of test files to include - 4890 is the full test set
+MAX_TEST_FILES = 4890
```
jorshi, line 32: Do we want to include the entire test set? Might make sense to provide easy comparison

heareval/tasks/speech_commands.py
```diff
@@ -82,50 +184,182 @@ def run(self):
...
-            pass
+        self.mark_complete()
 
 
-class SubsampleCorpus(SubsampleCorpus):
+class CombineMetaData(WorkTask):
+    """
+    Because we dealt with the metadata for the train/val and testing datasets
+    separately this combines them together into a single csv file.
+    """
```
khumairraj, line 270: `CombineMetaData` depends on the `ConfigureTestMetadata` and `ConfigureTrainValMetadata`. Mostly what the two-class for `test` and `trainval` are doing is to fetch the data for the splits followed by some operations which are independent of split(getting the `slug`, `label` and `file hash`). We can switch to the `ConfigureProcessMetadata` with the appropriate method to load these datasets and do the split independent tasks. I have worked on this class to this from one place. Will commit.

heareval/tasks/speech_commands.py
khumairraj: Yes. I think it is okay. We are doing a similar thing for the resampling as well, so this looks consistent solution in our script for running things in parallel.

khumairraj:
I have made a few changes and refactorings. Also if everything looks good, as next steps, will do the `coughvid` extractions with the classes defined in the `utils/luigi.py` - in a similar way as done in this `speechcommand` setup.

heareval/tasks/config/speech_commands.py
turian: Yeah why not

heareval/tasks/util/audio.py
turian: nice

heareval/tasks/util/audio.py
turian: Can you explain your comment?

turian (approved):
This is looking really nice!

heareval/tasks/util/audio.py
jorshi: Ya, my thought was that this function should remain a straightforward as possible and just run the resampling command when called. i.e, it shouldn't also be responsible for checking to make sure that it needs to resample. Part of the complication is that this function performs a resample and then saves the result in the new specified location (workdir) -- what would this function do if it doesn't need to resample? Copy the file to the new directory? Return an error? Create a symlink? 

If we want to avoid the redundancy of resampling an audio file that is already at the correct sample rate then the caller should be responsible for that. From the perspective of the luigi task it would totally make sense to skip calling this resample function and just create a symlink.

heareval/tasks/speech_commands.py
jorshi: Nice -- your refactors are excellent.

PR #74, opened 2021-07-20
Nsynth Pitch Dataset

jorshi:


heareval/tasks/nsynth_pitch.py
```diff
@@ -0,0 +1,225 @@
...
+        }
+
+    def run(self):
+        workdir = Path(self.workdir)
+        workdir.rmdir()
+        workdir.symlink_to(Path(self.requires()["train"].workdir).absolute())
+        self.mark_complete()
+
+
+class SplitTrainTestCorpus(luigi_util.SplitTrainTestCorpus):
```
jorshi, line 165: We can skip over the step that writes to mono and trims the length -- all nsynth samples are already all mono and exactly 64000 samples long (4s)

heareval/tasks/config/nsynth_pitch.py
```diff
@@ -0,0 +1,24 @@
...
+TEST_DOWNLOAD_URL = (
+    "http://download.magenta.tensorflow.org/datasets/nsynth/nsynth-test.jsonwav.tar.gz"
+)
+
+SAMPLE_LENGTH_SECONDS = 4.0
+
+# Number files to include
+MAX_TRAIN_FILES = 8000
+MAX_VAL_FILES = 1000
+MAX_TEST_FILES = 1000
```
jorshi, line 24: Again we can decide on what we want the full number of samples for the dataset to be here. Full nsynth is quite large, 289,205 train, 12,678 valid, and 4096 test. Probably want to use the full test set, just filter out pitches outside piano range. What do you think about the size for train / valid? A similar size to google speech command in terms of duration would be a bit over 20k samples for training.

jorshi:
mypy is complaining about some stuff

heareval/tasks/config/nsynth_pitch.py
turian: @jorshi I would include these details as a comment, but by default just use the entire set (None)

heareval/tasks/config/nsynth_pitch.py
```diff
@@ -0,0 +1,24 @@
+"""
+Configuration for the nsynth pitch detection task
+"""
+
+from .dataset_config import *  # noqa: F403, F401
+
+TASKNAME = "nsynth-pitch-v2.3.3"
+
```
turian, line 8: Weird there's no version number here

heareval/tasks/config/nsynth_pitch.py
```diff
@@ -0,0 +1,24 @@
+"""
+Configuration for the nsynth pitch detection task
+"""
```
turian, line 3: Could we maybe standardize these configs as like a JSON file (or Python dict) to make stuff more generic?

heareval/tasks/nsynth_pitch.py
```diff
@@ -0,0 +1,225 @@
...
+
+    outfile = luigi.Parameter()
+
+    def requires(self):
+        return {
+            "train": ExtractArchiveTrain(infile="train-corpus.tar.gz"),
+            "valid": ExtractArchiveValidation(infile="valid-corpus.tar.gz"),
+            "test": ExtractArchiveTest(infile="test-corpus.tar.gz"),
+        }
+
```
turian, line 64: These next two methods seem generic? But maybe it makes sense to keep them here?

heareval/tasks/nsynth_pitch.py
```diff
@@ -0,0 +1,225 @@
...
+    """
+    Aggregates subsampling of all the partitions into a single task as dependencies.
+    All the subsampled files are stored in the requires workdir, so we just link to
+    that since there aren't any real outputs associated with this task.
+    This is a bit of a hack -- but it allows us to avoid rewriting
+    the Subsample task as well as take advantage of Luigi concurrency.
+    """
+
+    def requires(self):
+        # Perform subsampling on each partition independently
```
turian, line 145: Yeah like here's an example of something that I think could be genericized if we had config as a dict? Or, maybe we don't even need that, just that the {partition: max_files} is a parameter

heareval/tasks/nsynth_pitch.py
turian: Add this as a comment?

I guess I would still prefer to include this step, because people might be monkey-patching based upon this code to create new pipelines. And my preferences would just be that if its already mono and trimmed, that the pipeline step is quite fast.

turian:
These are beginning to look quite nice.

heareval/tasks/nsynth_pitch.py
jorshi: Okay -- if we want to still include it, but make it fast if it is already correct, then I we will need an additional step in the MonoTrimWav task that checks each sample first and bypasses the conversion / trim if it is already correct. I think there is an ffmpeg function - ffprobe?

heareval/tasks/config/nsynth_pitch.py
jorshi: Sure sounds good. One thing to note is that running this on the entire NSynth dataset is going to take up alot of disk space. Between the workdir and task directory something like 700GB. So not sure how practical it will be for the average person to run this -- or for us to use that for the evaluation.

heareval/tasks/config/nsynth_pitch.py
jorshi: Ya let's add that as a separate config var?

heareval/tasks/config/nsynth_pitch.py
jorshi: Sure -- I'm into something like a json file

heareval/tasks/nsynth_pitch.py
jorshi: Ya, they are different than google speech though - so I think it makes sense to leave it

turian: @jorshi maybe because GSC has like postprocessing? that seems like it would be a common pattern

heareval/tasks/nsynth_pitch.py
```diff
@@ -0,0 +1,225 @@
...
+        logger.info(f"Preparing metadata for {split}")
+
+        # Loads and prepares the metadata for a specific split
+        split_path = Path(self.requires()[split].workdir).joinpath(f"nsynth-{split}")
+
+        metadata = pd.read_json(split_path.joinpath("examples.json"), orient="index")
+
+        # Filter out pitches that are not within the range of a standard piano
+        metadata = metadata[metadata["pitch"] >= 21]
+        metadata = metadata[metadata["pitch"] <= 108]
```
khumairraj, line 86: Should we define these ranges in the config file as well? Or maybe at the top of the file with uppercase variables.

jorshi: Ya good idea, let's move this into the config.

PR #80, opened 2021-07-22
Revert luigi to old pipeline

jorshi:
The current luigi pipeline in main is not working on Mac OSX. This was caused by a couple issues: 1) dynamically creating task classes did not work well for scheduling and pickling, 2) the task_name class attribute in WorkTask became a challenge when we moved to the runner.py method for invoking the pipelines.

We remove the dataset builder and add a new module called pipeline that has some pre-defined pipelines built using the task classes from util.luigi. This could potentially be refactored and cleaned up. We also simplify the config to just dicts.


heareval/tasks/nsynth_pitch.py
```diff
@@ -13,45 +13,28 @@
...
+        "train": "http://download.magenta.tensorflow.org/datasets/nsynth/nsynth-train.jsonwav.tar.gz",  # noqa: E501
+        "valid": "http://download.magenta.tensorflow.org/datasets/nsynth/nsynth-valid.jsonwav.tar.gz",  # noqa: E501
+        "test": "http://download.magenta.tensorflow.org/datasets/nsynth/nsynth-test.jsonwav.tar.gz",  # noqa: E501
+    },
+    "sample_duration": 4.0,
+    "partitions": [
+        {"name": "train", "max_files": 100},
+        {"name": "test", "max_files": 100},
+        {"name": "valid", "max_files": 100},
+    ],
```
turian, line 35: We intend to change these later

heareval/tasks/pipeline.py
```diff
@@ -0,0 +1,202 @@
...
+                max_files=partition["max_files"],
+                data_config=self.data_config,
+            )
+            for partition in self.data_config["partitions"]
+        }
+        return subsample_partitions
+
+    def run(self):
+        workdir = Path(self.workdir)
+        workdir.rmdir()
```
turian, line 76: Do we really need this rmdir? Can this cause bugs if some of the `SubsamplePartition` tasks complete successfully but then we run `SubsamplePartitions` again and it rmdirs?

turian (approved):

heareval/tasks/pipeline.py
jorshi: So the workdir in this task itself is a symlink to the previous requires workdir. The only time this is run is once all the  SubsamplePartition dependencies have completed - and it just creates a single symlink to the requires workdir (this basically just groups the multiple subsample dependencies into one). So the rmdir just removes a symlink if it exists, I think it is safe.

PR #87, opened 2021-07-23
DCASE2016 Task 2, HEAR 2021 version

turian:
This implements DCASE2016, Task 2.

The major change is that metadata now has a start and end column, for event-based labeling tasks.

Also:
* Each task has a task_type, event_labeling or scene_labeling
* partition is now called split everywhere

### DIFF
--- a/heareval/tasks/dcase2016_task2.py
+++ b/heareval/tasks/dcase2016_task2.py
@@ -42,41 +42,32 @@ config = {
 }
 
 
-class ExtractMetadata(luigi_util.WorkTask):
-    """
-    Custom metadata pre-processing. Creates a metadata csv
-    file that will be used by downstream luigi tasks to curate the final dataset.
-    TODO: It would be nice to have a better description of what this pattern is
-    """
-
-    outfile = luigi.Parameter()
+class ExtractMetadata(pipeline.ExtractMetadata):
     train = luigi.TaskParameter()
     test = luigi.TaskParameter()
 
     def requires(self):
+        return {"train": self.train, "test": self.test}
+
+    def split_to_path_str(self, split: str) -> str:
+        """
+        DCASE 2016 uses funny pathing, so we just hardcode the desired
+        (paths)
+        Note that for our training data, we only use DCASE 2016 dev data.
+        Their training data is short monophonic events.
+        """
         return {
-            "train": self.train,
-            "test": self.test,
-        }
+            "train": "dcase2016_task2_train_dev/dcase2016_task2_dev/",
+            "test": "dcase2016_task2_test_public/",
+        }[split]
 
-    @staticmethod
-    def get_rel_path(root: Path, item: pd.DataFrame) -> Path:
-        # Creates the relative path to an audio file given the note_str
-        audio_path = root.joinpath("audio")
-        filename = f"{item}.wav"
-        return audio_path.joinpath(filename)
-
-    @staticmethod
-    def slugify_file_name(filename: str) -> str:
-        return f"{slugify(filename)}.wav"
-
-    def get_split_metadata(self, split: str, split_path_str: str) -> pd.DataFrame:
+    def get_split_metadata(self, split: str) -> pd.DataFrame:
         logger.info(f"Preparing metadata for {split}")
 
         split_path = (
             Path(self.requires()[split].workdir)
             .joinpath(split)
-            .joinpath(split_path_str)
+            .joinpath(self.split_to_path_str[split])
         )
 
         metadatas = []
@@ -104,30 +95,6 @@ class ExtractMetadata(luigi_util.WorkTask):
             metadatas.append(metadata)
 
         return pd.concat(metadatas)[luigi_util.PROCESSMETADATACOLS]
-
-    def run(self):
-        # Get metadata for each of the data splits
-        process_metadata = pd.concat(
-            # DCASE 2016 uses funny pathing, so we just hardcode the desired
-            # paths
-            # Note that from our training data, we only use DCASE 2016 dev data.
-            # Their training data is short monophonic events.
-            [
-                self.get_split_metadata(
-                    "train", "dcase2016_task2_train_dev/dcase2016_task2_dev/"
-                ),
-                self.get_split_metadata("test", "dcase2016_task2_test_public/"),
-            ]
-        )
-
-        process_metadata.to_csv(
-            os.path.join(self.workdir, self.outfile),
-            columns=luigi_util.PROCESSMETADATACOLS,
-            header=False,
-            index=False,
-        )
-
-        self.mark_complete()
 
 
 def main(num_workers: int, sample_rates: List[int]):
--- a/heareval/tasks/dcase2016_task2.py
+++ b/heareval/tasks/dcase2016_task2.py
@@ -37,7 +37,6 @@ config = {
     "partitions": [
         {"name": "train", "max_files": 100},
         {"name": "test", "max_files": 100},
-        {"name": "valid", "max_files": 100},
     ],
 }
 
@@ -51,13 +50,11 @@ class ConfigureProcessMetaData(luigi_util.WorkTask):
     outfile = luigi.Parameter()
     train = luigi.TaskParameter()
     test = luigi.TaskParameter()
-    valid = luigi.TaskParameter()
 
     def requires(self):
         return {
             "train": self.train,
             "test": self.test,
-            "valid": self.valid,
         }
 
     @staticmethod
--- a/heareval/tasks/dcase2016_task2.py
+++ b/heareval/tasks/dcase2016_task2.py
@@ -10,7 +10,6 @@ We also allow training data outside this task.
 
 import logging
 import os
-from functools import partial
 from pathlib import Path
 from typing import List
 
--- a/heareval/tasks/dcase2016_task2.py
+++ b/heareval/tasks/dcase2016_task2.py
@@ -34,7 +34,7 @@ config = {
     # TODO: FIXME
     # Want different for train and test?
     "sample_duration": 120.0,
-    "partitions": [
+    "splits": [
         {"name": "train", "max_files": 10},
         {"name": "test", "max_files": 10},
     ],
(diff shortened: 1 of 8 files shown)


heareval/tasks/speech_commands.py
```diff
@@ -175,8 +175,10 @@ def get_metadata_attrs(self, process_metadata):
             # In this case we take the slug and remove the -nohash- as described
             # This nohash removal allows for the speech of similar person to be in the
             # same dataset. Such type of data specific requirements might be there.
+            # [what does this mean??]
             # in the readme of google speech commands. we want to keep similar people
             # in the same group - test or train or val
+            # [so do we do this or not?]
```
turian, line 181: Can someone explain what this is about? @jorshi @khumairraj ? I think we should talk this over.

heareval/tasks/util/luigi.py
```diff
@@ -17,10 +17,13 @@
 
 PROCESSMETADATACOLS = [
     "relpath",
+    # Like what is this and what is its contract?
```
turian, line 20: Can we please document ProcessMetadata and the contract behind its columns? There are several possible interpretations for what things are, how they are created, and what they mean? Could we just simplify this a lot?

heareval/tasks/util/luigi.py
```diff
@@ -176,15 +179,22 @@ def get_metadata(self):
...
 
         process_metadata = self.get_metadata()
         # Subsample the files based on max files per corpus.
         # The filename hash is used here
-        # This task can also be done in the configprocessmetadata as that will give
-        # freedom to stratify the selection on some criterion?
+        # This task can also be done in ConfigProcessMetadata as that will give
+        # freedom to stratify the selection on some criterion? That seems kinda
+        # fiddly, if we want fancy stratification we should discuss that
+        # separately.
```
turian, line 197: @jorshi @khumairraj I think we should talk this over.

heareval/tasks/dcase2016_task2.py
```diff
@@ -0,0 +1,144 @@
...
+
+config = {
+    "task_name": "dcase2016_task2",
+    "version": "hear2021",
+    "download_urls": {
+        "train": "https://archive.org/download/dcase2016_task2_train_dev/dcase2016_task2_train_dev.zip",  # noqa: E501
+        "test": "https://archive.org/download/dcase2016_task2_test_public/dcase2016_task2_test_public.zip",  # noqa: E501
+    },
+    # TODO: FIXME
+    # Want different for train and test?
```
jorshi, line 35: Ya -- this could be a partition specific config?

heareval/tasks/dcase2016_task2.py
```diff
@@ -0,0 +1,144 @@
...
+    def run(self):
+        # Get metadata for each of the data splits
+        process_metadata = pd.concat(
+            # DCASE 2016 uses funny pathing, so we just hardcode the desired
+            # paths
+            # Note that from our training data, we only use DCASE 2016 dev data.
+            # Their training data is short monophonic events.
+            [
+                self.get_split_metadata(
+                    "train", "dcase2016_task2_train_dev/dcase2016_task2_dev/"
```
jorshi, line 116: should we use os.path.join here to support windows as well?

heareval/tasks/speech_commands.py
jorshi: This maybe happens automatically with the hashing? But not sure -- @khumairraj ?

heareval/tasks/util/luigi.py
jorshi: Ya I was thinking about this. I just it might sense to have different metadata templates for each dataset type: scene classification / multi-label, timestamp multi-label, etc...

jorshi: One option would be to create a base case for each type that have the csv template. SceneBasedMetadata and TimestampBasedMetadata or something. Define the columns for each in that. Could also serve as a place to document the requirements for ProcessMetadata tasks.

heareval/tasks/util/luigi.py
jorshi: Ya -- we aren't doing any stratification presently. If we are subsampling severely we might need to think about it. Not sure exactly where to do that -- especially with sed tasks.

heareval/tasks/dcase2016_task2.py
turian: Whaddya mean?

heareval/tasks/speech_commands.py
turian: https://github.com/neuralaudio/hear2021-eval-kit/issues/93

heareval/tasks/speech_commands.py
```diff
@@ -175,8 +163,10 @@ def get_metadata_attrs(self, process_metadata):
             # In this case we take the slug and remove the -nohash- as described
             # This nohash removal allows for the speech of similar person to be in the
             # same dataset. Such type of data specific requirements might be there.
+            # [what does this mean??]
```
turian, line 166: https://github.com/neuralaudio/hear2021-eval-kit/issues/93

heareval/tasks/util/luigi.py
turian: https://github.com/neuralaudio/hear2021-eval-kit/issues/86

heareval/tasks/dcase2016_task2.py
```diff
@@ -0,0 +1,109 @@
...
+            )
+            assert os.path.exists(sound_file)
+            metadata = metadata.assign(relpath=sound_file)
+
+            metadata = metadata.assign(
+                slug=lambda df: df.relpath.apply(self.slugify_file_name)
+            )
+            metadata = metadata.assign(split=lambda df: split)
+            metadata = metadata.assign(
+                filename_hash=lambda df: df["slug"].apply(
```
turian, line 88: https://github.com/neuralaudio/hear2021-eval-kit/issues/88

heareval/tasks/pipeline.py
```diff
@@ -30,9 +32,94 @@ def get_download_and_extract_tasks(config: Dict):
...
+    We create a metadata csv file that will be used by downstream
+    luigi tasks to curate the final dataset.
+
+    The metadata columns are:
+        * relpath - How you find the file path in the original dataset.
+        * slug - This is the filename in our dataset. It should be
+        unique, it should be obvious what the original filename
+        was, and perhaps it should contain the label for audio scene
+        tasks.
+        * filename_hash - TODO: This should be slug_hash
```
turian, line 48: https://github.com/neuralaudio/hear2021-eval-kit/issues/88

heareval/tasks/speech_commands.py
```diff
@@ -140,10 +127,11 @@ def get_split_paths(self):
         val_silence = list(train_path.glob(f"{SILENCE}/running_tap*.wav"))
         validation_rel_paths.extend(val_silence)
         validation_df = pd.DataFrame(validation_rel_paths, columns=["relpath"]).assign(
-            partition=lambda df: "valid"
+            split=lambda df: "valid"
         )
 
         # Train files
+        # [really?? why is it called testing_list?]
```
turian, line 134: https://github.com/neuralaudio/hear2021-eval-kit/issues/93

heareval/tasks/util/luigi.py
```diff
@@ -321,15 +314,22 @@ def run(self):
...
+                # This won't work for sound event detection where there might be
+                # zero or more than one event per file
+                pass
+            else:
+                raise ValueError
+
+            # Save the slug and the label in as the split metadata
             sublabeldf.to_csv(
-                os.path.join(self.workdir, f"{partition}.csv"),
-                columns=["slug", "label"],
```
jorshi, line 330: if we want to keep these csv files clean we could use slug and label for scene_labeling and slug, label, start, and end for event_labeling -- and then save an additional metadata.csv file that is just the full set of all metadatas for all splits (including the split label).

jorshi (approved):

heareval/tasks/speech_commands.py
khumairraj: `.apply(lambda relpath: re.sub(r"-nohash-.*$", "", relpath))` This line removes the number and makes sure the hash of an audio file corresponds to the user id who recorded the example.
For example - `a_1, a_2, b_1, b_2`. If we hash the full name, the hash would not guarantee that the data by `a` and by `b` remains in different splits. This is a split condition to make sure the speech examples by the same person remains in the same split. Such a type of data specific split determination requires that we produce the hash in the metadata as it depends on the dataset. Such a requirement might not be there in another dataset and we might need to sample randomly(as is the case with others).

khumairraj: But since we are no more relying on the hash to decide the split(we are using the text split file with the examples), and rather just using the hash for subsampling, we can remove this. But still, this ensures, when we subsample, we subsample people recording the voice and not the data points. We can decide on one way or the other and remove it accrodingly.
`process_metadata = process_metadata.sort_values(by="filename_hash").iloc[:max_files]`

heareval/tasks/util/luigi.py
khumairraj: Yes. Since we were not doing any fancy preprocessing, I suggested to move this to the `ConfigProcessMetadata`. For speecheval we are subsampling the data on the basis of the id of the user recording the speech(handled by the filehash as discussed before). We can decide on keeping or moving this.

heareval/tasks/speech_commands.py
turian: @khumairraj Yes, I understand. I still think this is written in a confusing way and it mixes several concerns.

1) It's hard to understand from looking at the code what it is doing, and it would be nice to clean up the comments and make things more obvious.
2) The hash is potentially used for several things: a) Splitting into train / validation in a deterministic way. b) Subsampling a split in a determinstic way. Right now, with hash = person, b is broken because the behavior of instances for a particular person is not deterministic. We really need to think through these abstractions so that the semantics are obvious and clear.

The simplest thing might be have a separate split_hash and subsample_hash, and do splitting in Luigi for anything that doesn't have a validation set.

heareval/tasks/util/luigi.py
turian: @jorshi Agreed, I think we should have both.

PR #98, opened 2021-07-24
task embedder kinda works

turian:
task embedding kinda works, but it blows up on the long audio files from DCASE when batch_size > 1. We need to think this through

We can merge this for now, since we have a related issue #101

One thing we need to worry about is memory consumption. What happens when we want timestamp embeddings for a 10 minute audio? What is the maximum length audio we will provide in HEAR tasks? Is there a way that we can interrogate the memory usage of a model, without making too much extra work for the participants, and still appropriately choosing a batch size? How do we choose batches in a way that is flexible and maximally easy for participants?

Antoher thing we might consider is that the embedding .npy file should contain the labels for each embedding. That makes downstream predictor training easier (Does this work for ranking and JND? Let's talk this over.)


heareval/task_embeddings.py
```diff
@@ -22,102 +22,117 @@
...
 import numpy as np
+import pandas as pd
 import soundfile as sf
 import torch
 from torch.utils.data import DataLoader, Dataset
 from tqdm.auto import tqdm
 
 # This could instead be something from the participants
-EMBEDDING_PIP = "heareval.model.baseline"
+# TODO: Command line or config?
```
jorshi, line 39: Ya let's do this similar to hear validator? We can update to use click -- https://github.com/neuralaudio/hear-validator/blob/7d28d357620fff4c4a533e232a1079a1db785a49/hearvalidator/validate.py#L395 ?

heareval/task_embeddings.py
```diff
@@ -22,102 +22,117 @@
...
-        # TODO: We should be reading the metadata that describes
-        # the frame_rate.
-        # https://github.com/neuralaudio/hear2021-eval-kit/issues/53
-        frame_rate = 10
-
-        # TODO: Include "val" ?
-        for split in ["train", "test"]:
+        print(f"Task {task}")
+        task_config = json.loads(open(os.path.join(task, "task_metadata.json")).read())
+        for split in ["train", "valid", "test"]:
```
jorshi, line 113: Use the splits in the metadata? (task_config)

heareval/task_embeddings.py
```diff
@@ -22,102 +22,117 @@
...
-        for idx, row in enumerate(self.rows):
-            assert len(row) == ncol
-            self.rows[idx] = [row[0], row[1:]]
+        self.rows = pd.read_csv(csv_file)
+        # Since event detection metadata will have duplicates, we de-dup
+        # TODO: This suggests we might want to, instead of using CSV files,
+        # have a JSONL for metadata, with one file per line.
+        # Then there can be timestamp: list[labels].
+        # TODO: Make this a generic util function somewhere?
+        # TODO: Is this the right column to dedup on ???
```
jorshi, line 62: Why not, instead of use the task metadata, just look through all the audio files in the task directory and compute embeddings for them? Then we don't need to worry about all this reading of metadata and de-dupping stuff. We ultimately want embeddings for all the audio files so let's just use them and save the embeddings in the same folder structure with the same filename as the audios?

heareval/task_embeddings.py
jorshi: See comment below - perhaps we don't even need to be reading the metadata here.

heareval/task_embeddings.py
```diff
@@ -22,102 +22,117 @@
...
+            metadata_csv = os.path.join(task, f"{split}.csv")
+            if not os.path.exists(metadata_csv):
+                continue
+
             # TODO: We might consider skipping files that already
             # have embeddings on disk, for speed
+            # TODO: Choose batch size based upon audio file size?
+            # Or assume that the API is smart enough to do this?
+            # How do we test for memory blow up etc?
+            # e.g. that it won't explode on 10 minute audio
```
jorshi, line 125: Ya good question -- GPU memory management should be handled by the API right? Don't we demonstrate using a batch size to send to the model in a framewise way? Can we just load the audio into CPU memory and assume that the models are going to load them into GPU memory in a smart way?

turian: https://github.com/neuralaudio/hear-eval-kit/issues/101

jorshi (approved):

PR #106, opened 2021-07-24
[WIP] Task embeddings with labels

turian:


README.md
```diff
@@ -42,7 +46,13 @@ Options:
 
 
 
-[later we will include more details here]
+### Downstream Evaluation
+
+```
+python3 heareval/task_embeddings.py
```
jorshi, line 52: this won't work on a pip install. we want `python3 -m heareval.task_embeddings`

Should we create a separate 'runner' script to invoke generation of all the embeddings? That can accept the module and model weights from command line and run both task_embeddings and task_embeddings_with_labels

heareval/task_embeddings.py
```diff
@@ -93,6 +93,12 @@ def get_audio_embedding_numpy(
         )
         embeddings = embeddings.detach().cpu().numpy()
         timestamps = timestamps.detach().cpu().numpy()
+        # TODO: Turn this into a validator test
+        # What we really want is embeddings.shape[:2] == (batch_size, # timestamps) AND
+        # timestamps.shape[:2] == (batch_size, # timestamps)
+        assert len(embeddings) == len(
+            timestamps
+        ), f"{len(embeddings), embeddings.shape} != {len(timestamps)}"
```
jorshi, line 101: I'm on the fence about this -- I think having a shape of (n_timestamps, ) for the timestamps output is fine -- but if you feel strongly about the shape being (batch_size, n_timestamps) then let's do it. We'll update the baseline and validator to accommodate. Might be worth making this more explicit on the API documentation as well.

jorshi (approved):

PR #105, opened 2021-07-24
Fixing and Luigi Cleanup

khumairraj:
Tasks in the pull request:
- Change the slug in the metadata to not have a file extension since the preprocessing might change the extension(to `.wav`). Pick the extension after all the preprocessing has been done
- Move a dataset to the new pipeline format
- Work on a generic subsampling key. `(hash1, hash2, ..)`. Fix slugify file name for datasets.
- Cleaning up the worktask and other along with other tasks in the pipeline. Change to useing `pathlib` across pipeline.
- Moving the pipeline tasks from the `utils/luigi.py` to `pipeline.py` with descriptions added to most of them #81 
- Adding split level json for the different types of tasks

PR #112, opened 2021-07-26
[WIP] Task Embeddings Runner

jorshi:
Builds on the previous two PRs #98 and #106  for computing task embeddings and saving them + labels to npy files.

- Adds a runner class following the runner for the tasks. Invoked in a similar method. Decided to stick with click as opposed to hydra for simplicity.
- Combines the task_embeddings and task_labels functionality into the same module
-  Adding a Embedding class that wraps the imported embedding module to abstract calls to torch vs. tensorflow models
- Adds computing timestamp based label vectors for event detection based tasks.

Needs to add TensorFlow support. Also found a couple issues in the luigi pipeline to be addressed and then fixed here.


heareval/embeddings/runner.py
```diff
@@ -0,0 +1,47 @@
...
+@click.option(
+    "--tasks-dir",
+    default=None,
+    help="Location of tasks to compute embeddings on",
+    type=click.Path(exists=True),
+)
+def runner(module: str, model: str = None, tasks_dir: str = None) -> None:
+
+    # Check for directory containing the tasks
+    tasks_dir = Path("tasks") if tasks_dir is None else Path(tasks_dir)
```
turian, line 30: Why not just put that as the default in click?

turian: "tasks"

heareval/embeddings/task_embeddings.py
```diff
@@ -0,0 +1,279 @@
...
+                timestamp_labels[i, j, interval.data] = 1.0
+
+    return timestamp_labels
+
+
+def task_embeddings(embedding: Embedding, task_path: Path):
+
+    metadata = json.load(task_path.joinpath("task_metadata.json").open())
+
+    # FIXME: idx and label columns are switched in labelvocab
```
turian, line 233: Let's use headers

turian:
This is looking quite nice! Curious what is remaining in this PR

turian:
Update the README also?

heareval/embeddings/runner.py
jorshi: I experimented with that but the error message if the tasks dir doesn't exist isn't very helpful so I did it this way. I think the error message here could be even more descriptive and the help on the tasks-dir command line arg should also say that this will by default looks for a directory called "tasks" in the cwd.

heareval/embeddings/task_embeddings.py
jorshi: They are headers -- they are incorrect, needs to be fixed in the pipeline :)

jorshi:
@turian I think we want to finish and merge the luigi stuff #105 and update merge the baseline timestamp fixes and then test this before completing. I also want to go through and add (better) docstrings on everything and then update the readme for running. Would be cool to have a check for existence of embeddings for a task and not re-run them, but perhaps an additional PR.

turian (approved):

Issue #104, opened 2021-07-24
Statistics script for task embeddings


turian:
Summary statistics of task embedding output

Write a function for luigi that, for a particular directory, crawls it to find all WAV files and outputs stats to a JSON file:
* number of files
* mean, median, 25%, 75%, 10%, and 90% of audio file length

And then we want to run that both on the original WAV directories from extraction and then we want ALSO to run it at the end on the subsampled directories for each split. We will use this to sanity check stuff.

jorshi:
Completed in #123

PR #123, opened 2021-07-26
Luigi Stats

khumairraj:
Does the following 
- Makes an aggregate task `ResampleSubCorpuses` for the resampling of the audio files. This is done as this resample will also be used in the summary stats as well as the final task. This is similar to the `SubsampleSplits` aggregation.
- Make task `GenerateStats` for the extract and the sampled files. #104


heareval/tasks/pipeline.py
```diff
@@ -556,6 +557,119 @@ def run(self):
         self.mark_complete()
 
 
+class ResampleSubcorpuses(WorkTask):
+    """
+    Aggregates resampling of all the splits and sampling rates
+    into a single task as dependencies.
```
turian, line 563: Why? Won't this block and prevent parallelization? What's the benefit?

heareval/tasks/pipeline.py
```diff
@@ -575,18 +689,17 @@ class FinalizeCorpus(WorkTask):
...
-                ResampleSubCorpus(
-                    sr=sr,
-                    split=split,
-                    metadata=self.metadata,
-                    data_config=self.data_config,
-                )
-                for sr in self.sample_rates
-                for split in splits
-            ],
+            "stats": GenerateAudioStats(
```
turian, line 693: We want to do this on the original folders before ANY preprocessing also. I don't think this should be a task. It should just be a function you call within tasks.

heareval/tasks/util/audio.py
```diff
@@ -69,3 +71,12 @@ def resample_wav(in_file: str, out_file: str, out_sr: int):
     )
     # Make sure the return code is 0 and the command was successful.
     assert ret == 0
+
+
+def get_audiostats(in_file: str):
```
jorshi, line 76: Nice! We should include this in MonoWavTrimCorpus and only resample / convert to wav if the input needs it. If it doesn't then just create a symlink. Can do this in a separate PR.

heareval/tasks/pipeline.py
khumairraj: The benefit is that 
- We don't need to write the list of resampling tasks in the requirements of all other tasks after the `ResampleSubcorpus` and before the `FinalizeCorpus`. Since the `GenerateAudioStats` was something in the middle, I had to write the resample list in the requirements of both the `GenerateAudioStats` and `FinalizeCorpus`. 

Resample list which is referred above - 
https://github.com/neuralaudio/hear-eval-kit/blob/64fe04f24e4cd754495a95c127957c6d6504448a/heareval/tasks/pipeline.py#L579-L589

- This is consistent with how we are doing the `SubsampleSplits` which aggregate the `SubsampleSplit`. This will not affect parallelism as it is just a reference to the list of resampling tasks that will run in parallel.
https://github.com/neuralaudio/hear-eval-kit/blob/8365c0d2eac8b94b9b56af6b422087b6e33eab5b/heareval/tasks/pipeline.py#L286-L318

jorshi: It won't block parallelization, it basically just moves all separate requires tasks from finalizeCorpus out into a single task and aggregates the results into a single requires workdir for the GenerateAudioStats which simplifies that task. Is that correct @khumairraj ?

heareval/tasks/pipeline.py
```diff
@@ -575,18 +689,17 @@ class FinalizeCorpus(WorkTask):
...
+                metadata=self.metadata,
+                sample_rates=self.sample_rates,
+                data_config=self.data_config,
+            ),
+            "resample": ResampleSubcorpuses(
```
jorshi, line 698: Why have ResampleSubcorpuses a requirement here as well? It is already in the GenerateAudioStats so these will all necessarily be completed by the time we get here.

heareval/tasks/pipeline.py
```diff
@@ -556,6 +557,119 @@ def run(self):
...
+                for p in [10, 25, 75, 90]
+            }
+        )
+        return stats
+
+    def run(self):
+        master_stats = {"extracted": {}, "sampled": {}}
+        master_stats["extracted"] = self.get_stats(
+            self.get_metadata()["relpath"].apply(Path).tolist()
+        )
```
khumairraj, line 654: There are two parts to the stats 
- `extracted`: This is the summary of the list of `relpath` of audio files in the metadata file which has not been processed at all. `relpath` is a reference to all the extracted files(used this way because it helpes avoid looking out for different `extracted workdir` subfolders in a different dataset. The metadata always contain all the files with correct relative path)
- `sampled`: This is a dict and contains the stats for all the partitions and the sampling rates.

heareval/tasks/pipeline.py
jorshi: Ya agreed - I think we really want the report on the duration percentiles *before* we do any preprocessing and as a function that we call so we can get insight into a dataset and decide on what duration to set for MonoWavTrimCorpus, etc.. -- so this will be mega helpful prior to building out the full pipeline. 

This may be helpful as a sanity check at the end of the pipeline. But we would expect all the durations to be the same at this point.

We did discuss that it would be nice to get a stats report out detailing the number of audio files in each split and the different labels and distribution at the end of a pipeline ya?

khumairraj: Yes, We can remove it from being a task. We will have to track the different statistic JSON in different work folders though, to copy them in the final task(outside _workdir) output. In this task, we are just making one JSON, kind of the summary of what was the input and the output, and then copying this JSON over to the task dir(outside the `_workdir`). We can still go back to having this as a function. Please let me know your thoughts.

heareval/tasks/pipeline.py
```diff
@@ -556,6 +557,119 @@ def run(self):
...
+    Aggregates resampling of all the splits and sampling rates
+    into a single task as dependencies.
+
+    Parameter:
+        metadata (ExtractMetadata): task which extracts a corpus level metadata
+    Requires:
+        subsample_splits (list(SubsampleSplit)): task subsamples each split
+    """
+
+    sample_rates = luigi.Parameter()
```
jorshi, line 571: luigi.ListParameter()

jorshi: I don't think it really matters -- but it should be a list

heareval/tasks/pipeline.py
khumairraj: Yes.

heareval/tasks/util/audio.py
khumairraj: Yes. I was thinking of doing this. Will do.

heareval/tasks/pipeline.py
khumairraj: Because a) we might remove the GenerateAudioStats. b) We still require the workdirectory of the ResampleSubcorpuses to copy the final outputs

heareval/tasks/pipeline.py
khumairraj: Sure. Thanks

heareval/tasks/util/audio.py
jorshi: Awesome thanks

heareval/tasks/pipeline.py
khumairraj: But I get the requirement of setting the durations for the `MonoWavTrimCorpus`. We can make it a function and call in the extract metadata to save the stats in the same folder as the metadata. This is actually kind of a metadata of the input files in that case. We can keep the `GenerateAudioStats` in case we want to generate any other stats along with the once we are doing. Or we can remove as well.

jorshi: Ya -- I kind of like having this reported in the task dir as a summary of the input vs output -- but I think it's more useful from a debugging perspective, so maybe something we could optionally include as a task in the pipeline to help when building? @turian ?

A function that we can call on a folder of audio files that looks through recursively and reads the audio files and produces a report using the stats you have currently would be awesome.

heareval/tasks/pipeline.py
```diff
@@ -556,6 +557,119 @@ def run(self):
...
+        """Product the summary of a list of audio directory as dict"""
+        audio_stats = list(map(audio_util.get_audiostats, audiofiles))
+        durations = [stats["duration"] for stats in audio_stats]
+        sample_rates = [stats["sample_rate"] for stats in audio_stats]
+
+        stats = {
+            "total_audiofiles": len(durations),
+            "mean_samplerate": np.mean(sample_rates),
+            "mean_dur(sec)": np.mean(durations),
+            "median_dur)sec)": np.median(durations),
```
jorshi, line 640: mean samplerate isn't super meaningful -- I'd way rather have a list of samplerates to see which ones are present in a dataset. For the data before preprocessing having a list of formats would be cool too. There is def a chance that soundfile might not be able to read some of the formats that are present in the dataset though. (looking at you coughvid)

heareval/tasks/pipeline.py
turian: @khumairraj Again, we need the for datasets at the BEGINNING (before trim) and at the end. We want it at the beginning so we can actually pick appropriate audio file lengths for each particular task.

heareval/tasks/pipeline.py
khumairraj: Yes. This is a problem. In general, what will be a good way for the `audio_stats` function? Currently renamed to `audio_stats_wav` though. According to #104 . Maybe `ffmpeg` can help, but didnot get the stats directly for a `.webm` file. The following line - https://stackoverflow.com/questions/34118013/how-to-determine-webm-duration-using-ffprobe. The first comment talks about repackaging, please let me know if you already know any alternatives. Thanks.

heareval/tasks/util/audio.py
```diff
@@ -58,8 +65,8 @@ def resample_wav(in_file: str, out_file: str, out_sr: int):
             "-y",
             "-i",
             in_file,
-            "-af",
-            "aresample=resampler=soxr",
+            # "-af",
+            # "aresample=resampler=soxr",
```
jorshi, line 69: What's wrong with sox? Does this not work on your system?

heareval/tasks/util/audio.py
```diff
@@ -69,3 +76,39 @@ def resample_wav(in_file: str, out_file: str, out_sr: int):
...
+        "sample_rate": audio.samplerate,
+        "duration": round(len(audio) / audio.samplerate, 2),
+    }
+
+
+def audio_dir_stats_wav(in_dir: Union[str, Path], out_file: str):
+    """Produce summary by recursively searching a directory for wav files"""
+
+    audio_paths = list(Path(in_dir).absolute().rglob("*.wav"))
+    audio_dir_stats = list(map(audio_stats_wav, audio_paths))
```
jorshi, line 95: Is there a chance this will take a while to run on larger sets? If so a tqdm progress bar would be great.

jorshi (commented):
This is looking good! Would be awesome to be able to run the function audio_dir_stats_wav from the command line. Something like `python3 -m heareval.tasks.audio_dir_stats infile outfile ext`

heareval/tasks/util/audio.py
```diff
@@ -69,3 +76,39 @@ def resample_wav(in_file: str, out_file: str, out_sr: int):
...
+        "samples": len(audio),
+        "sample_rate": audio.samplerate,
+        "duration": round(len(audio) / audio.samplerate, 2),
+    }
+
+
+def audio_dir_stats_wav(in_dir: Union[str, Path], out_file: str):
+    """Produce summary by recursively searching a directory for wav files"""
+
+    audio_paths = list(Path(in_dir).absolute().rglob("*.wav"))
```
jorshi, line 94: Can we pass in the search extension as an arg to support other formats? Will need to be one that soundfile supports, but would be cool to be able to use this for other formats. Also perhaps a case insensitive search on the extension?

heareval/tasks/pipeline.py
turian: @khumairraj ^^

heareval/tasks/util/audio.py
```diff
@@ -69,3 +76,39 @@ def resample_wav(in_file: str, out_file: str, out_sr: int):
...
     assert ret == 0
+
+
+def audio_stats_wav(in_file: Union[str, Path]):
+    """Get statistics for a single wav file"""
+    audio = sf.SoundFile(in_file)
+    return {
+        "samples": len(audio),
+        "sample_rate": audio.samplerate,
+        "duration": round(len(audio) / audio.samplerate, 2),
```
turian, line 87: Why are we rounding here? Why not keep the precise value

heareval/tasks/util/audio.py
khumairraj: Yes. Some issue with configuring that in M1 mac. So I just remove the line, do the development and then uncomment.

khumairraj:
> This is looking good! Would be awesome to be able to run the function audio_dir_stats_wav from the command line. Something like `python3 -m heareval.tasks.audio_dir_stats infile outfile ext`

Sure. Have added this pushing in the following commit.

heareval/tasks/util/audio.py
khumairraj: Just to keep it approximate and good in the json. But will remove it.

heareval/tasks/util/audio.py
khumairraj: Yes. in the following commit.

README.md
```diff
@@ -44,6 +44,15 @@ Options:
                          default we resample to 16000, 22050, 44100, 48000.
 ```
 
+Additionally, to check the stats of an audio directory:
+```
+python3 -m heareval.tasks.audio_dir_stats {input folder} {output json file} {ext1} {ext2} ..
```
turian, line 49: Get rid of this and use .wav, .ogg, and .mp3 always

turian (approved):

Issue #124, opened 2021-07-26
Metadata JSON should be multilabel (list of labels)


turian:
Right now, it is always multiclass. Not flexible.

Issue #94, opened 2021-07-24
Create validation set in luigi?


turian:
The motivation for this is that, once embeddings are memmap'ed into a single file, it becomes hard to partition it by filename. In which case we might as well do it in Luigi earlier on.

Also, it just standardizes everything and makes it more likely everyone has the same dev set.

PR #125, opened 2021-07-26
memmap embeddings and labels

turian:
This will write embeddings to a memmap'ed file and labels to a single pickle'd file. This will make it easier to iterate over them quickly when learning the downstream predictor.

A few gotchas come up:
* The label format is inconsistent and weird between scenes and timestamps. Also scene labels are multiclass (one label) not lists. This isn't flexible. #124 I mean maybe this is okay tho, because for multiclass we want to return a tensor that is a single long and for multilabel we want to return a one hot tensor. (We can't have variable length rows in a tensor.)
* Once the files are memmap'ed it's slow to partition the examples into train and validation. I think luigi should split upstream into train and validation set for consistency. #94

### DIFF
--- a/heareval/embeddings/task_embeddings.py
+++ b/heareval/embeddings/task_embeddings.py
@@ -23,6 +23,8 @@ TODO:
 """
 import json
 import os.path
+import pickle
+import random
 import shutil
 from importlib import import_module
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


heareval/embeddings/task_embeddings.py
```diff
@@ -217,14 +219,97 @@ def get_labels_for_timestamps(labels: List, timestamps: np.ndarray) -> List:
...
+    for embedding_file in tqdm(embedding_files):
+        emb = np.load(embedding_file)
+        lbl = json.load(
+            open(str(embedding_file).replace("embedding.npy", "target-labels.json"))
+        )
+        if metadata["embedding_type"] == "scene":
+            assert emb.ndim == 1
+            embedding_memmap[idx] = emb
+            # This is janky AF. The format is inconsistent with event embeddings
+            # AND its multiclass not multilabel. Honestly this should just be a list of labels.
```
jorshi, line 287: Agreed -- list of labels is great. What is inconsistent about this the event embedding format? They need to be different no? Or just that it's not a list of labels.

jorshi (approved):
This is cool and I think is fine for now. But it feels a bit clunky to be writing things twice. Especially when these single files could be massive. I can't think of a better solution for the timestamp based files atm though.

heareval/embeddings/task_embeddings.py
turian: I just don't think scene should be a dict and event should be a list.

turian:
@jorshi We could just write it once to the memmap'ed file and not to the individual files. Shouldn't be too hard.

https://github.com/neuralaudio/hear-eval-kit/issues/128

PR #127, opened 2021-07-26
Randomized predictions

turian:
Simple randomized prediction (no learning or anything) from embeddings.


heareval/predictions/task_predictions.py
```diff
@@ -0,0 +1,127 @@
...
+    # Only do test for now, for the unlearned random predictor
+    for split in [{"name": "test"}]:
+        print(f"Getting embeddings for split: {split['name']}")
+
+        dataloader = DataLoader(
+            SplitMemmapDataset(embedding_path, split["name"]),
+            batch_size=64,
+            # We don't shuffle because it's slow.
+            # Also we want predicted labels in the same order as
+            # target labels.
```
jorshi, line 111: Is the issue that reading adjacent items using memmap is significantly faster? Also -- I think if we using the dataset to return target labels then we don't need to worry about the ordering here.

heareval/predictions/task_predictions.py
```diff
@@ -0,0 +1,127 @@
...
+    embedding_path: Path, scene_embedding_size: int, timestamp_embedding_size: int
+):
+    metadata = json.load(embedding_path.joinpath("task_metadata.json").open())
+
+    label_vocab = pd.read_csv(embedding_path.joinpath("labelvocabulary.csv"))
+
+    nlabels = len(label_vocab)
+    assert nlabels == label_vocab["idx"].max() + 1
+
+    if metadata["embedding_type"] == "scene":
```
jorshi, line 90: Nice -- the addition of the embedding_type and prediction_type here makes the model selection here clean. Is there a chance we are going to have task-dependent predictors as well (i.e. choose a different model for nsynth vs speech commands even though they are the same embedding and prediction type)? Or try to use the same architecture for most of the tasks?

heareval/predictions/runner.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python3
+"""
+Downstream training, using embeddings as input features and learning
+predictions.
+
+TODO: Add CUDA device.
```
jorshi, line 6: This can just be a click option ya?

jorshi (approved):
Nice -- this is looking great.

heareval/predictions/runner.py
```diff
@@ -0,0 +1,63 @@
...
+@click.option(
+    "--model",
+    default=None,
+    help="Location of model weights file",
+    type=click.Path(exists=True),
+)
+def runner(
+    module: str, embeddings_dir: str = "embeddings", model: Optional[str] = None
+) -> None:
+    embeddings_dir = Path(embeddings_dir).joinpath(module)
```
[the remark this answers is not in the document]
turian: @jorshi Where? Can you show me? Sure I'm in favor of consisteny.

heareval/predictions/task_predictions.py
turian: @jorshi we're going to train in this order too.

Random access to a memmap is slow because you're seeking around disk

heareval/predictions/task_predictions.py
turian: We will really try to use the same architecture everywhere

heareval/predictions/runner.py
turian: Agreed yeah

turian: @jorshi I mean same for task_embedding.py, you should be able to specify the device on the command-line

turian: https://github.com/neuralaudio/hear-eval-kit/issues/129

heareval/predictions/runner.py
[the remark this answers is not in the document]
jorshi: https://github.com/neuralaudio/hear-eval-kit/blob/20b7c4b4a16afd5951d9fba88289fc65c10feaf1/heareval/embeddings/task_embeddings.py#L83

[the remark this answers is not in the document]
jorshi: So that's after importing the module. Could do something like:
```
module = import_module(module)
embeddings_dir = Path(embeddings_dir).joinpath(module.__name__)
```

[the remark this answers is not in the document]
turian: @jorshi maybe let's just use the string everywhere? Because then we don't actually need to load code necessarily just to get the directory name.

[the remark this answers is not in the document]
jorshi: Sure that's a good idea. Let's do that. Let's update the task_embeddings then.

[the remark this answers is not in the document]
jorshi: #139

PR #141, opened 2021-07-27
List for scene classes

jorshi:
Closes https://github.com/neuralaudio/hear-eval-kit/issues/124

### DIFF
--- a/heareval/tasks/pipeline.py
+++ b/heareval/tasks/pipeline.py
@@ -340,7 +340,7 @@ class MonoWavTrimCorpus(WorkTask):
     def run(self):
         # TODO: this should check to see if the audio is already a mono wav at the
         #   correct length and just create a symlink if that is this case.
-        for audiofile in tqdm(self.requires()["corpus"].workdir.iterdir()):
+        for audiofile in tqdm(list(self.requires()["corpus"].workdir.iterdir())):
             newaudiofile = self.workdir.joinpath(f"{audiofile.stem}.wav")
             audio_util.mono_wav_and_fix_duration(
                 audiofile, newaudiofile, duration=self.data_config["sample_duration"]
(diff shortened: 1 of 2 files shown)


turian:
Okay but shouldn't we be checking in a few places, e.g. in the luigi pipelines and in task_prediction or task_embedding that multiclass tasks have exactly one label per embedding?

heareval/tasks/pipeline.py
```diff
@@ -340,7 +340,7 @@ def requires(self):
     def run(self):
         # TODO: this should check to see if the audio is already a mono wav at the
         #   correct length and just create a symlink if that is this case.
-        for audiofile in tqdm(self.requires()["corpus"].workdir.iterdir()):
+        for audiofile in tqdm(list(self.requires()["corpus"].workdir.iterdir())):
```
jorshi, line 343: Quick fix for tqdm progress bar

heareval/tasks/pipeline.py
```diff
@@ -441,12 +441,20 @@ def run(self):
...
+                    assert len(audiolabel_df) == len(audiodf)
+
+                # Create a dictionary containing a list of labels keyed on the slug.
                 audiolabel_json = (
                     audiolabel_df[["slug_path", "label"]]
-                    .set_index("slug_path")
-                    .to_dict("index")
+                    .groupby("slug_path")["label"]
+                    .apply(list)
+                    .to_dict()
```
jorshi, line 457: Before this was:
```
{'slug_name': {'label': 'dog'}, ...}
```

Updated to be:
```
{'slug_name': ["dog"], ...}
```
This is to be more consistent with event based when it comes time to doing embeddings.

heareval/tasks/pipeline.py
```diff
@@ -441,12 +441,20 @@ def run(self):
             )
 
             if self.data_config["embedding_type"] == "scene":
-                # For scene labeling each scene should have one
-                assert len(audiolabel_df) == len(audiodf)
```
turian, line 445: Honestly this is the wrong place to do this. We should be doing this as early in the pipeline as possible. Why not do this in ExtractMetadata so that the metadata format is consistent through the entire pipeline, rather than changing halfway through?

jorshi: Sure, that sounds like a good place to perform the check. Let me see if there is a logical place to insert it.

jorshi: Done -- moved up to to ExtractMetadata

turian (approved):

Issue #142, opened 2021-07-28
Copy slice of memmapped array to memory before converting to memory?


jorshi:
Getting this warning in task_prediction: 
```
home/jshier/anaconda3/lib/python3.8/site-packages/torch/utils/data/_utils/collate.py:64: UserWarning: The given NumPy array is not writeable, and PyTorch does not support non-writeable tensors. This means you can write to the
underlying (supposedly non-writeable) NumPy array using the tensor. You may want to copy the array to protect its data or make it writeable before converting it to a tensor. This type of warning will be suppressed for the rest
 of this program. (Triggered internally at  /pytorch/torch/csrc/utils/tensor_numpy.cpp:180.)
  return default_collate([torch.as_tensor(b) for b in batch])
```

Potential solution:
```
def __getitem__(self, idx):
    return np.copy(self.embedding_memmap[idx])
```

turian:
Solved be solved in https://github.com/neuralaudio/hear-eval-kit/pull/167

PR #143, opened 2021-07-28
[WIP] Evaluation

jorshi:
Starting on the evaluation runner. Metric for a task are defined in the task metadata and are automatically run for each embedding / task.


heareval/predictions/task_predictions.py
```diff
@@ -75,7 +75,7 @@ def __len__(self):
         # return len(self.labels)
 
     def __getitem__(self, idx):
-        return self.embedding_memmap[idx]
+        return np.copy(self.embedding_memmap[idx])
```
jorshi, line 78: #142 -- added this to try. Can keep it if we want. Probably needs to be profiled.

heareval/evaluation/task_evaluation.py
```diff
@@ -0,0 +1,94 @@
...
+
+def task_evaluation(task_path: Path):
+
+    metadata = json.load(task_path.joinpath("task_metadata.json").open())
+    label_vocab = pd.read_csv(task_path.joinpath("labelvocabulary.csv"))
+
+    embedding_type = metadata["embedding_type"]
+
+    if "evaluation" not in metadata:
+        print(f"Task {task_path.name} has no evaluation config.")
```
jorshi, line 61: This should raise an error

heareval/evaluation/runner.py
```diff
@@ -0,0 +1,49 @@
...
+from heareval.evaluation.task_evaluation import task_evaluation
+
+
+@click.command()
+@click.option(
+    "--embeddings-dir",
+    default="embeddings",
+    help="Location of task embeddings to compute evaluation on.",
+    type=click.Path(exists=True),
+)
```
turian, line 21: Wait why don't we need the package we're evaluating here? hearbaseline etc.

jorshi: Why not compute evaluation on all them at once? They are all subfolders in the embeddings dir. Can add an option to specify and just run one.

heareval/evaluation/task_evaluation.py
```diff
@@ -10,91 +10,123 @@
...
+        for i, prediction in enumerate(predictions):
+            predicted_class = np.argmax(prediction)
+            assert len(targets[i]) == 1
+            target_class = label_vocab[targets[i][0]]
+
+            if predicted_class == target_class:
+                correct += 1
+
+        error = correct / len(targets)
+        return {"top1_error": error}
```
khumairraj, line 74: `sklearn.metrics` has a `top_k_accuracy_score`. We can use that `k==1`

jorshi: I like the idea of using the sklearn metric

heareval/evaluation/runner.py
turian: True.

jorshi:
I think this can be closed once #165 is merged?

turian:
#165 pulled this in

Issue #107, opened 2021-07-24
Filename hash => sort key


turian:
An alterative would be to create a field called: "sort key" which would be like a tuple of hashes. Then we could use it both for subsample and for splitting. It could be like: "(person hash, filename hash)"

turian:
For dataset splitting, we have the keys that cannot be broken across multiple splits. e.g. person or instrument should all be in one split.

Let's call this "split hash"

PR #148, opened 2021-07-28
Luigi Subsample Keys

khumairraj:
Add a split key and a subsample key and sort by split key and then by subsample key to get stable subsampling for the three datasets. #107


heareval/tasks/pipeline.py
```diff
@@ -154,34 +155,35 @@ def slugify_file_name(relative_path: str):
...
+        Gets the split key.
+
+        This can be the hash of a data specific split like instrument generating
+        the file or person recording the sound.
+        For subsampling, the data is subsampled in chunks of split.
+        By default this is the hash of the filename, but can be overridden if the
+        data needs to split into certain groups while subsampling
+        i.e leave out some groups and select others based on this key.
+        """
+        assert "relpath" in df, "relpath column not found in the dataframe"
```
jorshi, line 169: Nice I didn't know about assert messages -- that's good to know

heareval/tasks/pipeline.py
```diff
@@ -282,7 +293,9 @@ def run(self):
             print(f"{num_files} audio files in corpus, keeping only {max_files}")
 
         # Sort by the subsample key and select the max_files number of samples
-        metadata = metadata.sort_values(by="subsample_key").iloc[:max_files]
+        metadata = metadata.sort_values(
+            by=["split_key", "subsample_key"], ascending=[True, True]
+        ).iloc[:max_files]
```
jorshi, line 298: Should we issue a warning here if the set doesn't contain examples covering all the labels? I don't think we want to raise an error since it is useful to have small splits for testing and debugging, but if all the labels aren't present then downstream evaluation will break (mostly just AUC).

jorshi (approved):

jorshi:
This looks good. We had talked about trying to use sklearn to do stratified subsampling -- but I think that won't be necessary when we are using the full test set. Can probably be included in a separate PR if we decide to go that route.

turian (approved):

turian:
Agree on @jorshi 's feedback, specifically we have issues if all labels are not in every split, but I think that could be a subsequent PR

PR #149, opened 2021-07-28
Event prediction

jorshi:
For event predictions this converts a set of frame based label predictions to a list of events with start and stop timestamps in ms.

### DIFF
--- a/heareval/predictions/task_predictions.py
+++ b/heareval/predictions/task_predictions.py
@@ -94,6 +94,8 @@ def create_events_from_prediction(
 
     # Apply thresholding to get the label detected at each frame.
     # TODO: Apply median filtering for smoothing?
+    #   Additionally, DCASE limits the number of active classes at a single
+    #   time to 6. We could do that here as well.
     predictions = (predictions > threshold).type(torch.int8)
 
     # Difference between each timestamp to find event boundaries
@@ -128,7 +130,9 @@ def create_events_from_prediction(
     return event_tree
 
 
-def get_predictions_as_events(predictions: torch.Tensor, file_timestamps: List):
+def get_predictions_as_events(
+    predictions: torch.Tensor, file_timestamps: List, label_vocab: pd.DataFrame
+) -> Dict[str, List]:
     # This probably could be more efficient if we make the assumption that
     # each frame in the prediction and timestamps are in order.
     assert predictions.shape[0] == len(file_timestamps)
@@ -144,11 +148,26 @@ def get_predictions_as_events(predictions: torch.Tensor, file_timestamps: List):
         # Save the predictions for the file keyed on the timestamp
         event_files[slug][timestamp] = predictions[i]
 
-    event_trees = {}
-    for slug, timestamp_predictions in tqdm(event_files.items()):
-        event_trees[slug] = create_events_from_prediction(timestamp_predictions)
+    # Dictionary of labels: {idx -> label}
+    label_dict = label_vocab.set_index("idx").to_dict()["label"]
 
-    return event_trees
+    # Create events for all the different files. Store all the events as a dictionary
+    # with the same format as the ground truth from the luigi pipeline.
+    # { slug -> [{"label" : "woof", "start": 0.0, "end": 2.32}, ...], ...}
+    event_dict = {}
+    for slug, timestamp_predictions in tqdm(event_files.items()):
+        event_tree = create_events_from_prediction(timestamp_predictions)
+        events = []
+        for interval in sorted(event_tree):
+            label = label_dict[interval.data]
+            # TODO: start and end? Let's use begin and end like interval tree?
+            events.append(
+                {"label": label, "start": interval.begin, "end": interval.end}
+            )
+
+        event_dict[slug] = events
+
+    return event_dict
 
 
 def task_predictions(
@@ -200,9 +219,19 @@ def task_predictions(
                     f"{split['name']}.filename-timestamps.json"
                 ).open()
             )
-            events = get_predictions_as_events(all_predicted_labels, file_timestamps)
-            print(events)
-            assert False
+            print("Creating events from predictions:")
+            events = get_predictions_as_events(
+                all_predicted_labels, file_timestamps, label_vocab
+            )
+            annotation_dir = embedding_path.joinpath(
+                "events", f"{split['name']}-prediction"
+            )
+            annotation_dir.mkdir(parents=True, exist_ok=True)
+            json.dump(
+                events,
+                embedding_path.joinpath(f"{split['name']}.predictions.json").open("w"),
+                indent=4,
+            )
 
         pickle.dump(
             all_predicted_labels,
--- a/heareval/predictions/task_predictions.py
+++ b/heareval/predictions/task_predictions.py
@@ -14,10 +14,13 @@ TODO:
     many models simultaneously with one disk read?
 """
 
+from collections import OrderedDict
 import json
 import pickle
 from pathlib import Path
+from typing import Dict, List
 
+from intervaltree import IntervalTree
 import numpy as np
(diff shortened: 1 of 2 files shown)


heareval/predictions/task_predictions.py
```diff
@@ -30,8 +32,9 @@ def __init__(self, nfeatures: int, nlabels: int, prediction_type: str):
         super().__init__()
 
         self.projection = torch.nn.Linear(nfeatures, nlabels)
+        torch.nn.init.normal_(self.projection.weight)
```
jorshi, line 35: Was experimenting with different inits for testing -- can remove this.

heareval/predictions/task_predictions.py
```diff
@@ -120,9 +234,34 @@ def task_predictions(
             predicted_labels = predictor(embs)
             # TODO: Uses less memory to stack them one at a time
             all_predicted_labels.append(predicted_labels)
-        all_predicted_labels = torch.cat(all_predicted_labels)
+        predicted_labels = torch.cat(all_predicted_labels)
```
jorshi, line 237: mypy

heareval/predictions/task_predictions.py
```diff
@@ -79,6 +82,117 @@ def __getitem__(self, idx):
...
+            # New onset
+            if pred_diff[t][class_idx] == 1:
+                assert len(current_event) == 0
+                current_event = [timestamps[t]]
+
+            # Offset for current event
+            elif pred_diff[t][class_idx] == -1:
+                assert len(current_event) == 1
+                # TODO: filter on event length? DCASE only included events that
+                #   are 60ms long.
```
jorshi, line 133: **atleast** 60ms

heareval/predictions/task_predictions.py
turian: okay, I'm fine either way

heareval/predictions/task_predictions.py
```diff
@@ -79,6 +82,117 @@ def __getitem__(self, idx):
         # return self.embedding_memmap[idx], self.labels[idx]
```
turian, line 84: On the one hand, I'm not sure we want to do this here. On the other hand, I have no idea where we want to do this? So I guess this is fine :)

heareval/predictions/task_predictions.py
turian: This is worth a discussion :\

heareval/predictions/task_predictions.py
```diff
@@ -79,6 +82,117 @@ def __getitem__(self, idx):
...
+    # TODO: Apply median filtering for smoothing?
+    #   Additionally, DCASE limits the number of active classes at a single
+    #   time to 6. We could do that here as well.
+    predictions = (predictions > threshold).type(torch.int8)
+
+    # Difference between each timestamp to find event boundaries
+    pred_diff = torch.diff(predictions, dim=0)
+
+    # This is slow, but for each class look through all the timestamp predictions
+    # and construct a set of events with onset and offset timestamps.
```
turian, line 114: I really am not sure what this is doing, is it what `more_itertools.consecutive_groups` does? https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.consecutive_groups

heareval/predictions/task_predictions.py
```diff
@@ -79,6 +82,117 @@ def __getitem__(self, idx):
         # return self.embedding_memmap[idx], self.labels[idx]
 
 
+def create_events_from_prediction(
+    prediction_dict: Dict, threshold: float = 0.5
```
turian, line 86: Maybe since we expect the key to be float timestamp, etc. put that in the docstring and type here?

turian: Like Dict[float, float]

turian: I still am not 100% sure what the signature is for prediction dict. Does it involve the label or index? Is there are probability?

heareval/predictions/task_predictions.py
jorshi: It's based on the dcase code - it's basically looking for boundaries between consecutive groups of the *same* element. It achieves that by using the difference function above. So you get a value of 1 when you have an onset and then -1 for an onset because we have converted to binary vectors. So it starts at the first timestamp and loops through looking for the boundaries and adds an event when there has been an onset and then an offset.

heareval/predictions/task_predictions.py
jorshi: Ya -- could be defined for each task

heareval/predictions/task_predictions.py
jorshi: Updated this : Dict[float, torch.Tensor]

heareval/predictions/task_predictions.py
```diff
@@ -79,6 +83,133 @@ def __getitem__(self, idx):
...
 
 
+def create_events_from_prediction(
+    prediction_dict: Dict[float, torch.Tensor],
+    threshold: float = 0.5,
+    median_filter_ms=90.0,
+    min_duration=60.0,
+) -> IntervalTree:
+    """
+    Takes a set of prediction tensors keyed on timestamps and generates events.
```
jorshi, line 93: Better decsription

heareval/predictions/task_predictions.py
```diff
@@ -79,6 +83,133 @@ def __getitem__(self, idx):
...
+    timestamps = np.array(sorted(prediction_dict.keys()))
+
+    # Create a sorted numpy matrix of frame level predictions for this file. We convert
+    # to a numpy array here before applying a median filter.
+    predictions = np.stack([prediction_dict[t].detach().numpy() for t in timestamps])
+
+    # Apply median filtering and then thresholding to get the
+    # label detected at each frame.
+    # TODO: DCASE limit the number of concurrent events to 5 - should we?
+    ts_diff = timestamps[1] - timestamps[0]
```
jorshi, line 117: Check that all timestamps are the same

heareval/predictions/task_predictions.py
```diff
@@ -79,6 +83,133 @@ def __getitem__(self, idx):
...
+    filter_width = int(round(median_filter_ms / ts_diff))
+
+    # Apply filter along the frames for each class to smooth out the labels
+    predictions = median_filter(predictions, size=(filter_width, 1))
+    predictions = (predictions > threshold).astype(np.int8)
```
jorshi, line 122: Add a comment -- binary vector

heareval/predictions/task_predictions.py
```diff
@@ -79,6 +83,133 @@ def __getitem__(self, idx):
...
+    predictions = np.stack([prediction_dict[t].detach().numpy() for t in timestamps])
+
+    # Apply median filtering and then thresholding to get the
+    # label detected at each frame.
+    # TODO: DCASE limit the number of concurrent events to 5 - should we?
+    ts_diff = timestamps[1] - timestamps[0]
+    filter_width = int(round(median_filter_ms / ts_diff))
+
+    # Apply filter along the frames for each class to smooth out the labels
+    predictions = median_filter(predictions, size=(filter_width, 1))
```
jorshi, line 121: Remove filter -- can give an unfair advantage depending on timestamp interval

heareval/predictions/task_predictions.py
```diff
@@ -79,6 +83,133 @@ def __getitem__(self, idx):
...
+                if event_duration >= min_duration:
+                    event_tree.addi(
+                        begin=current_event[0], end=current_event[1], data=class_idx
+                    )
+                current_event = []
+
+    return event_tree
+
+
+def get_predictions_as_events(
```
jorshi, line 159: Better names or more detailed comments

turian (approved):

PR #150, opened 2021-07-29
Event eval

jorshi:
Sound event detection metrics from sed_eval. Implements both segment-based and event-based metrics that were used in DCASE2016 Task2. They used segment-based total error and onset-only event-based f1. This returns all segment-based and event-based metrics so we may need to filter these -- perhaps a task-config var?

Depends on the event predictions saved as JSON files in #149

### DIFF
--- a/heareval/evaluation/task_evaluation.py
+++ b/heareval/evaluation/task_evaluation.py
@@ -79,7 +79,7 @@ class MacroAUC(MetricFunction):
     def __call__(
         self, predictions: np.ndarray, targets: List, **kwargs
     ) -> Dict[str, float]:
-        return {"auc": None}
+        # return {"auc": None}
         # Dictionary of labels and integer idx: {label -> idx}
         label_vocab = self.label_vocab_as_dict(key="label")
 
--- a/heareval/evaluation/task_evaluation.py
+++ b/heareval/evaluation/task_evaluation.py
@@ -7,13 +7,16 @@ TODO: This is a start on this file. Still need to add a bunch of metrics for the
     Would it make sense to move all the metric functions over to a file called metrics?
 """
 
+from functools import partial
 import json
 from pathlib import Path
 import pickle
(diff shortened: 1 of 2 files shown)


heareval/evaluation/task_evaluation.py
```diff
@@ -96,19 +175,50 @@ def task_evaluation(task_path: Path):
...
+    embedding_type = metadata["embedding_type"]
+
+    if "evaluation" not in metadata:
+        print(f"Task {task_path.name} has no evaluation config.")
+        return
+
+    if embedding_type == "scene":
+        predictions, targets = get_scene_based_prediction_files(task_path)
+    elif embedding_type == "event":
+        predictions, targets = get_event_based_prediction_files(task_path)
```
jorshi, line 213: This is getting a bit nasty with the event detection embeddings dir having a couple different types of labels files. Both the frame-based labels as well as the event-based labels. Perhaps in another PR we should try to clean that up a bit -- or organize the files produced in the event detection type embeddings directories.

heareval/evaluation/task_evaluation.py
```diff
@@ -63,22 +66,98 @@ def macroauc(
...
-    if "evaluation" not in metadata:
-        print(f"Task {task_path.name} has no evaluation config.")
-        return
+# TODO: Add additional metrics
 
+available_metrics: Dict[str, Callable] = {
+    "top1_error": top1_error,
+    "macroauc": macroauc,
+    "event_based": event_based_metrics,
+    "onset_only_event_based": partial(event_based_metrics, offset=False),
```
jorshi, line 155: We might want to have specific values (like tolerance) depend on a config variable -- we could include an optional kwargs dict for evaluation metrics in the task_metadata to support that if we anticipate needing to set different metric parameters more easily.

turian (approved):

PR #152, opened 2021-08-01
Setup Stratification in Luigi Task

khumairraj:
This adds stratify key to stratify the subsampling process. This is a key that is by default set to be the label key. However, the stratification behaviour can be controlled by overriding `get_stratify_key`. The method is based on a first grouping by the stratify key followed by sorting each group by the split key, subsample key and then selecting the required fraction of data points from each group.


heareval/tasks/pipeline.py
```diff
@@ -284,18 +317,49 @@ def get_metadata(self):
...
+            by=["split_key", "subsample_key"], ascending=[True, True]
+        )
+        # Groupby and select the required sample from each group
+        sampled_metadata = pd.concat(
+            [
+                # Ensure at least 1 sample is selected for each group
+                stratify_grp.head(1 if grp_count[grp] < 1 else int(grp_count[grp]))
+                for grp, stratify_grp in metadata.groupby("stratify_key")
+            ]
+        )
```
khumairraj, line 335: Process - 
- Get the count required for each group.
- Sort the metadata by the `split_key` and the `subsample_key`
- For each group select the count required and concatanate

heareval/tasks/pipeline.py
```diff
@@ -284,18 +317,49 @@ def get_metadata(self):
...
+        # Do some assertions
+        assert set(sampled_metadata["stratify_key"].unique()) == set(
+            metadata["stratify_key"].unique()
+        ), "All stratify groups are not in the sampled metadata."
+        # Add the num of groups here for handling addition in case of highly
+        # imbalanced subsampling
+        assert len(sampled_metadata) <= max_files + len(
+            grp_count
+        ), "Sampled metadata is more than the allowed max files + unique groups"
```
khumairraj, line 344: Checks
- if all the labels are there in the method
- if the subsampled data points are more than the `max_subsample + len(grp_count)`. The length of group count is here since some groups might be too small and we might need to take one sample for the group.

heareval/tasks/pipeline.py
```diff
@@ -284,18 +317,49 @@ def get_metadata(self):
...
-            by=["split_key", "subsample_key"], ascending=[True, True]
-        ).iloc[:max_files]
+            print(
+                f"{num_files} audio files in corpus."
+                "Max files to subsample: {max_files}"
+            )
+            metadata = self.subsample_metadata(metadata, max_files)
+            assert self.subsample_metadata(metadata, max_files).equals(
+                metadata
+            ), "The subsampling is not stable"
```
khumairraj, line 362: Check
- Is the subsampling stable across two runs?

heareval/tasks/pipeline.py
```diff
@@ -154,6 +154,23 @@ def slugify_file_name(relative_path: str):
...
+        Since hashing is only required for ordering the samples
+        for subsampling, the stratify key should not necessarily be a hash,
+        as it is only used to group the data points before subsampling.
+        The actual subsampling is done by the split key and
+        the subsample key.
+
+        By default, the label is used for stratification
+        """
+        assert "label" in df, "label column not found in the dataframe"
+        return df["label"].apply(str).apply(filename_to_int_hash)
```
turian, line 172: a) isn't this already a string? Why do we need to apply str?
b) Why are we hashing this?

heareval/tasks/pipeline.py
```diff
@@ -239,6 +265,13 @@ def run(self):
             index=False,
         )
 
+        for split, split_df in process_metadata.groupby("split"):
```
turian, line 268: What is this doing? Is it redundant?

heareval/tasks/pipeline.py
```diff
@@ -284,18 +317,49 @@ def get_metadata(self):
         )
         return metadata
 
+    def subsample_metadata(self, metadata, max_files):
+
+        # Get group count for each group
+        grp_count = metadata["stratify_key"].value_counts() * max_files / len(metadata)
```
turian, line 323: why not do int(round(...))

heareval/tasks/pipeline.py
turian: This should be a comment

heareval/tasks/pipeline.py
turian: Same, this should be a comment

heareval/tasks/pipeline.py
turian: Okay, should this also be a unit test?

heareval/tasks/pipeline.py
turian: maybe even

max(1, int(round(...))) so it's always more than zero

turian:
This looks pretty good but:
a) it's a little confusing and needs some comments. Maybe also even subroutines.
b) can we write unit tests for the subrountines??

heareval/tasks/pipeline.py
khumairraj: It is grouping by the split to see count for each group

heareval/tasks/pipeline.py
khumairraj: Will remove. Just kept for sanity. Also it need not be a hash as stated in the comments. Will change that as well

heareval/tasks/pipeline.py
khumairraj: Yes. we can include it in the unit test for the pipeline. I have done some fixes and made it more robust for now here.

turian (approved):
Very nice.

Issue #157, opened 2021-08-05
nsynth metadata columns are missing


auroracramer:
[This assertion](https://github.com/neuralaudio/hear-eval-kit/blob/main/heareval/tasks/pipeline.py#L252) fails when running `python -m heareval.tasks.runner nsynth_pitch`:

```
Traceback (most recent call last):
  File "/ext3/miniconda3/lib/python3.9/site-packages/luigi/worker.py", line 191, in run
    new_deps = self._run_get_new_deps()
  File "/ext3/miniconda3/lib/python3.9/site-packages/luigi/worker.py", line 133, in _run_get_new_deps
    task_gen = self.task.run()
  File "/ext3/code/hear-eval-kit/heareval/tasks/pipeline.py", line 242, in run
    assert set(
AssertionError
```

It may be due to the last line of [nsynth_pitch.ExtractMetadata.get_split_metadata](https://github.com/neuralaudio/hear-eval-kit/blob/ca210b9e3b52fec4cc5f37aa8c2a38374ffa8d0f/heareval/tasks/nsynth_pitch.py#L100) omitting the `split_key` and `stratify_keys` expected when processing metadata for scene embeddings.

khumairraj:
Thank you @auroracramer for bringing this up. 
The solution is in fact to just return all the columns in the last line [nsynth_pitch.ExtractMetadata.get_split_metadata](https://github.com/neuralaudio/hear-eval-kit/blob/ca210b9e3b52fec4cc5f37aa8c2a38374ffa8d0f/heareval/tasks/nsynth_pitch.py#L100) by changing the line to be just  `return metadata`

The nsynth data is indeed very large and I couldn't get beyond the download task. However, I am working on a smaller version of all datasets that could help us with setting up CI/CD. I will fix this issue in the same pr and also run end to end for nsynth.

khumairraj:
#160 fixes this.

turian:
@auroracramer closing the issue but please comment or open another issue as you need

PR #160, opened 2021-08-05
Sample dataset Task

khumairraj:
Builds a sampler for the dataset
- Uses the same download and extract tasks to make sure the same downloaded files can be used for sampling
- Also uses the configs defined in the task files for making it simple to scale across multiple datasets

### DIFF
--- a/heareval/tasks/nsynth_pitch.py
+++ b/heareval/tasks/nsynth_pitch.py
@@ -97,7 +97,7 @@ class ExtractMetadata(pipeline.ExtractMetadata):
             )
         )
 
-        return metadata[["relpath", "slug", "subsample_key", "split", "label"]]
+        return metadata
 
 
 def main(num_workers: int, sample_rates: List[int]):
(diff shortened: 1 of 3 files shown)


turian (commented):
Please update the README to describe when this should be used and by whom (i.e. just devs as a one time thing, and that the downsamples should go into https://github.com/turian/hear2021-open-tasks-downsampled)

turian (approved):

heareval/tasks/nsynth_pitch.py
```diff
@@ -97,7 +97,7 @@ def get_split_metadata(self, split: str) -> pd.DataFrame:
             )
         )
 
-        return metadata[["relpath", "slug", "subsample_key", "split", "label"]]
```
khumairraj, line 100: This line fixes the Nsynth Pitch Bug. It returns the data frame with all the columns. #157

heareval/tasks/nsynth_pitch.py
```diff
@@ -97,7 +97,7 @@ def get_split_metadata(self, split: str) -> pd.DataFrame:
...
+        return metadata
```
khumairraj, line 100: Nsynth Fix. This will return all the columns now. #157

turian (approved):

PR #162, opened 2021-08-11
Add sampled dataset

khumairraj:
Adds
- download the small version of the data from [HEAR OPEN TASK](https://github.com/turian/hear2021-open-tasks-downsampled)
- add `-small` tag to the version of the task and corresponding change in the runner for running the small task.
- add option to change the base directory for the task and the luigi intermediate outputs.
- fixes the subsampling to be deterministic by changing the subsample key to be hash of the slug.
- cleanup the sampler and make it deterministic as well

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,13 @@ mean meadian and certain (10, 25, 75, 90) percentile durations.
 This is helpful in getting a quick glance of the audio files in a folder and 
 helps in decideing the preprocessing configurations.
 
+For development, there is a sampler which could be used to generate small corpuses for testing of each task
+```
+python -m heareval.tasks.sampler <taskname>
+```
+Supported task name are speech_commands, nsynth_pitch and dcase2016_task2
+Sampled small version of the tasks can be found here: [Downsampled HEAR Open Tasks](https://github.com/turian/hear2021-open-tasks-downsampled)
+
 ### Computing embeddings
 
 Once a set of tasks has been generated, embeddings can be computed using any audio
--- a/README.md
+++ b/README.md
@@ -35,8 +35,9 @@ For DCASE 2016, Task 2 (sound event detection):
 python3 -m heareval.tasks.runner dcase2016_task2
 ```
 
-These commands will download and preprocess the entire dataset. An intermediary dir
-call `_workdir` will be created, and then a final directory called `tasks` will contain
+These commands will download and preprocess the entire dataset. An intermediary directory
+defined by the option `luigi-dir`(default `_workdir`) will be created, and then a 
+final directory defined by the option `tasks-dir` (default `tasks`) will contain
 the completed dataset.
 
 Options:
@@ -50,6 +51,12 @@ Options:
   
   --small       FLAG     If passed, the task will run on a small-version of the 
                          data.
+
+  --luigi-dir   STRING   Path to dir to store the intermediate luigi task outputs.
+                         By default this is set to _workdir in the module root directory
+
+  --tasks-dir   STRING   Path to dir to store the final task outputs.
+                         By default this is set to tasks in the module root directory
 ```
 The small flag runs the preprocessing pipeline on a small version of each dataset stored at [Downsampled HEAR Open Tasks](https://github.com/turian/hear2021-open-tasks-downsampled). This is used for development and continuous integration tests for the pipeline. These small versions of the data can be generated deterministically with the following command:
 ```
--- a/README.md
+++ b/README.md
@@ -48,10 +48,14 @@ Options:
   --sample-rate INTEGER  Perform resampling only to this sample rate. By
                          default we resample to 16000, 22050, 44100, 48000.
   
-  --small       BOOL     If True, the task will run on a small-version of the 
+  --small       FLAG     If passed, the task will run on a small-version of the 
                          data.
 ```
-[Link to small version of data](https://github.com/turian/hear2021-open-tasks-downsampled)
+The small flag runs the preprocessing pipeline on a small version of each dataset stored at [Downsampled HEAR Open Tasks](https://github.com/turian/hear2021-open-tasks-downsampled). This is used for development and continuous integration tests for the pipeline. These small versions of the data can be generated deterministically with the following command:
+```
+python -m heareval.tasks.sampler <taskname>
+```
+Supported task name are speech_commands, nsynth_pitch and dcase2016_task2.
 
 Additionally, to check the stats of an audio directory:
 ```
@@ -61,13 +65,6 @@ Stats include: audio_count, audio_samplerate_count,
 mean meadian and certain (10, 25, 75, 90) percentile durations.
 This is helpful in getting a quick glance of the audio files in a folder and 
 helps in decideing the preprocessing configurations.
-
-For development, there is a sampler which could be used to generate small corpuses for testing of each task
-```
-python -m heareval.tasks.sampler <taskname>
-```
-Supported task name are speech_commands, nsynth_pitch and dcase2016_task2
-Sampled small version of the tasks can be found here: [Downsampled HEAR Open Tasks](https://github.com/turian/hear2021-open-tasks-downsampled)
 
 ### Computing embeddings
 
--- a/README.md
+++ b/README.md
@@ -47,6 +47,8 @@ Options:
 
   --sample-rate INTEGER  Perform resampling only to this sample rate. By
                          default we resample to 16000, 22050, 44100, 48000.
+  
+  --small BOOL If True, the task will run on a small [Version](https://github.com/turian/hear2021-open-tasks-downsampled) of the data
 ```
 
 Additionally, to check the stats of an audio directory:
--- a/README.md
+++ b/README.md
@@ -48,8 +48,10 @@ Options:
   --sample-rate INTEGER  Perform resampling only to this sample rate. By
                          default we resample to 16000, 22050, 44100, 48000.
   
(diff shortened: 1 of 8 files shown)


heareval/tasks/dcase2016_task2.py
```diff
@@ -88,7 +87,9 @@ def get_split_metadata(self, split: str) -> pd.DataFrame:
                 .replace("annotation", "sound")
                 .replace(".txt", ".wav")
             )
-            assert os.path.exists(sound_file)
+            # Remove the assert statement as this file might not exist.
+            # This is anyways ensured in the Base ExtractMetadata task.
+            # assert os.path.exists(sound_file)
```
khumairraj, line 92: Removed this `assert` because, in the case of sampled data, files in the metadata might not exist in the folder. This check has been moved in the `run` part of the `ExtractMetadata` task. If files are not found, the count of files not found will be printed and those rows will be removed from the metadata so that there is no error while subsampling from the small dataset.

heareval/tasks/sampler.py
```diff
@@ -45,7 +45,7 @@
     },
     "dcase2016_task2": {
         "task_config": dcase2016_task2_config,
-        "audio_sample_size": 10,
+        "audio_sample_size": 4,
```
khumairraj, line 48: Decrease the number of files in the audio sample size for dcase dataset so that the file is less than 25 mb so that it can be in the github repo.

heareval/tasks/util/luigi.py
```diff
@@ -121,7 +121,9 @@ def download_file(url, local_filename, expected_md5):
                 f.write(chunk)
                 pbar.update(chunk_size)
             pbar.close()
-    assert md5sum(local_filename) == expected_md5
+    # If expected_md5 is 0, the check is not done
+    if expected_md5:
+        assert md5sum(local_filename) == expected_md5
```
khumairraj, line 126: If md5 is an empty string, the md5 will not be checked. Since the dataset for downloading is small enough in small tasks we are not checking the md5.

turian: Can we still use the small file's md5sum ?? then we can verify everything is correct

any way we can easily check also that the small file has a certain # of instances and certain # of labels in the final postprocessed version?

README.md
```diff
@@ -47,7 +47,11 @@ Options:
 
   --sample-rate INTEGER  Perform resampling only to this sample rate. By
                          default we resample to 16000, 22050, 44100, 48000.
+  
+  --small       BOOL     If True, the task will run on a small-version of the 
+                         data.
```
turian, line 52: False is the default if this is unspecified right?

Can you just make this argument a flag, either it exists or it doesn't, and no bool value is needed?

heareval/tasks/pipeline.py
```diff
@@ -261,6 +286,18 @@ def run(self):
                 "%s embedding_type unknown" % self.data_config["embedding_type"]
             )
 
+        # Remove files from the process metadata which donot exist in the dataset.
```
turian, line 289: spelling

heareval/tasks/dcase2016_task2.py
```diff
@@ -88,7 +87,9 @@ def get_split_metadata(self, split: str) -> pd.DataFrame:
                 .replace("annotation", "sound")
                 .replace(".txt", ".wav")
             )
-            assert os.path.exists(sound_file)
+            # Remove the assert statement as this file might not exist.
```
turian, line 90: Is there a way to know if small mode is enabled? In which case we can run this when small is disabled.

heareval/tasks/pipeline.py
turian: Is there a way to know if small mode is enabled? In which case we can run this only when small is enabled.

heareval/tasks/dcase2016_task2.py
khumairraj: Yes. There is a way. But in general, we don't check this for all the tasks. I can add a general check in the pipeline for this. The check will run when small is false.

heareval/tasks/dcase2016_task2.py
```diff
@@ -44,7 +48,29 @@
     "splits": [
         {"name": "train", "max_files": 10},
         {"name": "test", "max_files": 10},
+        {"name": "valid", "max_files": 10},
```
jorshi, line 51: Are these the actual split max_files that we are going to use during evaluation? Have we set the final values for these yet for all the open tasks or are we still in dev mode for this?

jorshi (commented):
Nice -- overall looking cool! Made a few comments and have some clarifying questions.

heareval/tasks/dcase2016_task2.py
khumairraj: These are still in dev.

heareval/tasks/speech_commands.py
```diff
@@ -207,8 +228,18 @@ def get_process_metadata(self) -> pd.DataFrame:
...
+def main(
+    num_workers: int,
+    sample_rates: List[int],
+    luigi_dir: str,
+    tasks_dir: str,
+    small: bool = False,
+):
+    if small:
+        config.update(config["small"])
+    config.update(**{"luigi_dir": luigi_dir})
```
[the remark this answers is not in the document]
khumairraj: Actually some mypy issue without the star. I was resolving it so added this. Will change

heareval/tasks/dcase2016_task2.py
turian: @khumairraj Can we just set max_files to None in the non-small case, so I can switch to doing full preprocessing of the open tasks?

heareval/tasks/sampler.py
```diff
@@ -84,25 +88,27 @@ def sample(self, all_files):
...
+            )
+            .assign(
+                stratify_key=lambda df: df.audio_path.apply(lambda path: path.parent),
+                split_key=lambda df: df.audio_path.apply(
+                    lambda path: luigi_util.filename_to_int_hash(str(path))
+                ),
+                subsample_key=lambda df: df.split_key,
+            )
+            .pipe(luigi_util.subsample_metadata, self.audio_sample_size)
+            .audio_path.to_list()
```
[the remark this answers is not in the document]
khumairraj: Splitted and added comments as well.

heareval/tasks/util/luigi.py
khumairraj: Changed to using the md5 for the small dataset as well

README.md
```diff
@@ -47,7 +47,11 @@ Options:
...
 ```
+[Link to small version of data](https://github.com/turian/hear2021-open-tasks-downsampled)
```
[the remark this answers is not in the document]
khumairraj: Changed in the readme to add more specific description.

heareval/tasks/dcase2016_task2.py
```diff
@@ -44,7 +48,29 @@
...
+                "url": "https://github.com/turian/hear2021-open-tasks-downsampled/raw/main/dcase2016_task2_train_dev-small.zip",  # noqa: E501
+                "md5": "3adca7e1860aedf7d1b0c06358f1b867",
+            },
+            {
+                "name": "test",
+                "url": "https://github.com/turian/hear2021-open-tasks-downsampled/raw/main/dcase2016_task2_test_public-small.zip",  # noqa: E501
+                "md5": "73446e2156cb5f1d44a1de1e70e536a5",
+            },
+        ],
+        "small_flag": True,
```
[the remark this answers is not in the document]
khumairraj: Planned to use the flag -- when we check whether all the files in the metadata were found in the download corpus. But I have removed this flag as this is redundant and can be checked from the version itself -- version has a `-small` attatched at the end.

heareval/tasks/pipeline.py
```diff
@@ -261,6 +261,22 @@ def run(self):
...
+        if sum(exists) < len(process_metadata):
+            if "small" in self.data_config:
+                print(
+                    "All Files in metadata doesnot exist in the dataset. "
+                    f"Removing {len(process_metadata) - sum(exists)} entries in the "
+                    "metadata"
+                )
+                process_metadata = process_metadata.loc[exists]
+            else:
+                raise FileNotFoundError("All files in metadata are not found")
```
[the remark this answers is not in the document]
khumairraj: Added comments for that.

heareval/tasks/dcase2016_task2.py
khumairraj: Will handle in a different pr as discussed.

turian (approved):

heareval/tasks/pipeline.py
```diff
@@ -193,36 +202,81 @@ def get_subsample_key(df: DataFrame) -> Series:
...
+        # test and valid if not found in the metadata can be sampled
+        # from the train
+        assert "train" in splits_present, "Train split not found in metadata"
+        splits_to_sample = set(["test", "valid"]).difference(splits_present)
+        print(f"Splits getting sampled with the split key are: {splits_to_sample}")
+
+        # Depending on whether valid and test are already present, the percentage can
+        # either be the predefined percentage or 0
+        valid_perc = VALIDATION_PERCENTAGE if "valid" in splits_to_sample else 0
+        test_perc = VALIDATION_PERCENTAGE if "test" in splits_to_sample else 0
```
jorshi, line 263: Is this supposed to be TESTING_PERCENTAGE?

khumairraj: Thanks Jordie

PR #165, opened 2021-08-13
Split predict into train/test

turian:
We still use the unlearned predictor, which we simply return from the train method.

However, now we do label mapping and properly return labels for multiclass and multilabel cases.


heareval/predictions/task_predictions.py
```diff
@@ -68,18 +82,37 @@ def __init__(self, embedding_path: Path, split_name: str):
...
-        # return len(self.labels)
 
-    def __getitem__(self, idx):
-        return self.embedding_memmap[idx]
-        # return self.embedding_memmap[idx], self.labels[idx]
+    def __getitem__(self, idx) -> Tuple[torch.Tensor, torch.Tensor]:
+        x = self.embedding_memmap[idx]
+        y = [self.label_to_idx[str(label)] for label in self.labels[idx]]
+        if self.prediction_type == "multilabel":
+            # Lame special case
```
jorshi, line 100: Special case being a sample with zero labels?

heareval/predictions/task_predictions.py
```diff
@@ -235,14 +273,35 @@ def task_predictions(
...
+    predictor: torch.nn.Module,
+    embedding_path: Path,
+    scene_embedding_size: int,
+    timestamp_embedding_size: int,
+):
+    metadata = json.load(embedding_path.joinpath("task_metadata.json").open())
+    label_vocab, nlabels = label_vocab_nlabels(embedding_path)
+
+    if metadata["embedding_type"] == "scene":
+        predictor = RandomProjectionPrediction(
```
jorshi, line 289: Don't we want to use the passed in predictor in this function?

heareval/predictions/task_predictions.py
```diff
@@ -41,6 +42,7 @@ def __init__(self, nfeatures: int, nlabels: int, prediction_type: str):
             raise ValueError(f"Unknown prediction_type {prediction_type}")
 
     def forward(self, x: torch.Tensor):
+        x, y = x
```
jorshi, line 45: What's `y` for?

jorshi: Wait, so is the x that is passed in here a tuple that contains the embeddings and the ground truth? That's quite confusing, can we split that out before this?

heareval/predictions/task_predictions.py
```diff
@@ -92,14 +125,15 @@ def create_events_from_prediction(
     converts the prediction tensor to a binary label based on the threshold value. Any
     events occurring at adjacent timestamps are considered to be part of the same event.
     This loops through and creates events for each label class. Disregards events that
-    are less than the min_duration.
+    are less than the min_duration milliseconds.
```
jorshi, line 128: nice

heareval/predictions/task_predictions.py
turian: yes

heareval/predictions/task_predictions.py
turian: No the passed in predictor was used to create *embeddings*. We learn our own predictor on top. But we afterwards will change this to single hidden layer.

heareval/predictions/task_predictions.py
turian: I fix this in a later PR

heareval/predictions/task_predictions.py
```diff
@@ -35,28 +49,76 @@ def __init__(self, nfeatures: int, nlabels: int, prediction_type: str):
         torch.nn.init.normal_(self.projection.weight)
         if prediction_type == "multilabel":
             self.activation: torch.nn.Module = torch.nn.Sigmoid()
+            self.logitloss = torch.nn.BCEWithLogitsLoss()
```
khumairraj, line 52: I think this should be `self.logit_loss`

heareval/predictions/task_predictions.py
```diff
@@ -35,28 +49,76 @@ def __init__(self, nfeatures: int, nlabels: int, prediction_type: str):
...
         elif prediction_type == "multiclass":
             self.activation = torch.nn.Softmax()
+            self.logit_loss = OneHotToCrossEntropyLoss()
```
khumairraj, line 55: Do we need `torch.nn.Softmax()` as activation? We already are using `torch.nn.CrossEntropyLoss()` in the `OneHotToCrossEntropyLoss()`, which does softmax before applying the cross entropy?

khumairraj (approved):

heareval/predictions/task_predictions.py
turian: Done

heareval/predictions/task_predictions.py
turian: You can see now this is fixed

heareval/predictions/task_predictions.py
jorshi: I'm actually still a bit confused by this -- you pass in the predictor as a arg to this function and then overwrite it immediately? What's the point in passing in the predictor?

heareval/predictions/task_predictions.py
```diff
@@ -215,44 +296,77 @@ def get_events_for_all_files(
...
         )
     elif metadata["embedding_type"] == "event":
-        predictor = RandomProjectionPrediction(
+        predictor = PredictionModel(
             timestamp_embedding_size, nlabels, metadata["prediction_type"]
         )
     else:
         raise ValueError(f"embedding_type {metadata['embedding_type']} unknown.")
+    # train on CPU
+    # TODO: FIXME
```
jorshi, line 339: What's broken? Can you expand on this comment a bit?

heareval/evaluation/task_evaluation.py
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env python3
+"""
+Contains a set of various evaluation scores. Runs evaluation on the test predictions
+for predictions on a HEAR task.
+Would it make sense to move all the score functions over to a file called scores?
+
+TODO: Fix AUC
```
jorshi, line 7: Update docstring

jorshi (approved):

PR #167, opened 2021-08-13
Learn the predictor

turian:
Unfortunately, we skip dcase because it doesn't have a validation set. https://github.com/neuralaudio/hear-eval-kit/issues/94

### DIFF
--- a/heareval/predictions/task_predictions.py
+++ b/heareval/predictions/task_predictions.py
@@ -20,12 +20,25 @@ import pickle
 from pathlib import Path
 from typing import Dict, List, Tuple
 
-from intervaltree import IntervalTree
 import numpy as np
 import pandas as pd
+import pytorch_lightning as pl
 import torch
-from torch.utils.data import Dataset, DataLoader
+from intervaltree import IntervalTree
+from torch.utils.data import DataLoader, Dataset
 from tqdm.auto import tqdm
+
+
+class OneHotToCrossEntropyLoss(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+        self.loss = torch.nn.CrossEntropyLoss()
+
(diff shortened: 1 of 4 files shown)


heareval/predictions/task_predictions.py
```diff
@@ -36,30 +49,65 @@ def __init__(self, nfeatures: int, nlabels: int, prediction_type: str):
...
 
 
+class PredictionModel(pl.LightningModule):
+    def __init__(self, nfeatures: int, nlabels: int, prediction_type: str):
+        super().__init__()
+
+        self.predictor = RandomProjectionPrediction(nfeatures, nlabels, prediction_type)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        return self.predictor(x)
```
jorshi, line 76: So is the activation function only applied during testing?

heareval/predictions/task_predictions.py
```diff
@@ -257,22 +304,43 @@ def label_vocab_nlabels(embedding_path: Path) -> Tuple[pd.DataFrame, int]:
...
         )
     elif metadata["embedding_type"] == "event":
-        predictor = RandomProjectionPrediction(
+        predictor = PredictionModel(
             timestamp_embedding_size, nlabels, metadata["prediction_type"]
         )
     else:
         raise ValueError(f"embedding_type {metadata['embedding_type']} unknown.")
+    # train on CPU
+    # TODO: FIXME
```
jorshi, line 339: What's broken?

jorshi (approved):
Looks cool!

heareval/predictions/task_predictions.py
turian: yes, but it is in forward too

heareval/predictions/task_predictions.py
turian: just that we want to specify more parameters here, like device

PR #169, opened 2021-08-13
Use downstream evaluation measures during validation

turian:
This came together quite nicely because of the `ScoreFunction` class. So nice thinking ahead, it was very clean.

This pulls in #143 and I think we can merge it? And do the sed_eval in a subsequent PR.


heareval/score.py
```diff
@@ -0,0 +1,189 @@
...
+            self.name = name
+
+    def __call__(self, predictions: Any, targets: Any, **kwargs) -> float:
+        """
+        Compute the score based on the predictions and targets. Returns the score.
+        """
+        raise NotImplementedError("Inheriting classes must implement this function")
+
+    def __str__(self):
+        return self.name
```
jorshi, line 53: This looks like it will fail for the event detection methods below that don't have names defined. If name is optional and potentially not defined, do you want to return the name of the class if name isn't a defined attribute?

heareval/score.py
```diff
@@ -0,0 +1,189 @@
...
+    See https://tut-arg.github.io/sed_eval/generated/sed_eval.sound_event.EventBasedScores.html # noqa: E501
+    for params.
+    """
+
+    score_class = sed_eval.sound_event.EventBasedMetrics
+
+
+available_scores: Dict[str, Callable] = {
+    "top1_err": Top1Error,
+    "pitch_err": partial(Top1Error, name="pitch_err"),
```
jorshi, line 183: Nice -- I like the name arg for renaming a metric

heareval/evaluation/task_evaluation.py
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env python3
+"""
+Contains a set of various evaluation scores. Runs evaluation on the test predictions
+for predictions on a HEAR task.
+Would it make sense to move all the score functions over to a file called scores?
+
+TODO: Fix AUC
```
jorshi, line 7: Is AUC still broken?

heareval/evaluation/task_evaluation.py
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env python3
+"""
+Contains a set of various evaluation scores. Runs evaluation on the test predictions
+for predictions on a HEAR task.
+Would it make sense to move all the score functions over to a file called scores?
```
jorshi, line 5: Let's update this docstring, looks like this was done here

heareval/predictions/task_predictions.py
```diff
@@ -89,10 +102,11 @@ def validation_step(self, batch, batch_idx):
...
         y_hat = self.predictor.forward_logit(x)
         loss = self.predictor.logit_loss(y_hat, y)
+        y_pr = self.predictor(x)
         # Logging to TensorBoard by default
-        self.log("val_loss", loss)
-        # TODO: Use the actual evaluation criterion?
-        return loss
+        self.log("val_loss", loss, prog_bar=True)
+        for score in self.scores:
+            self.log(f"val_{score}", score(y_pr, y), prog_bar=True)
```
jorshi, line 109: Should this still return a loss? I see that you're logging the validation now -- what do we use for early stopping now?

heareval/predictions/task_predictions.py
```diff
@@ -320,96 +342,117 @@ def dataloader_from_split_name(
...
+        embedding_size, label_to_idx, nlabels, metadata["prediction_type"], scores
+    )
+
+    # First score is the target
+    target_score = f"val_{str(scores[0])}"
+
+    checkpoint_callback = ModelCheckpoint(monitor=target_score, mode="max")
+    early_stop_callback = EarlyStopping(
+        monitor=target_score, min_delta=0.00, patience=10, verbose=False, mode="max"
+    )
```
jorshi, line 362: Oh cool, this is telling lightning to monitor the first score and use it for early stopping? Which is why validation step doesn't need to return anything?

heareval/predictions/task_predictions.py
```diff
@@ -320,96 +342,117 @@ def dataloader_from_split_name(
...
-            scene_embedding_size, nlabels, metadata["prediction_type"]
-        )
-    elif metadata["embedding_type"] == "event":
-        predictor = PredictionModel(
-            timestamp_embedding_size, nlabels, metadata["prediction_type"]
-        )
-    else:
-        raise ValueError(f"embedding_type {metadata['embedding_type']} unknown.")
     # train on CPU
     # TODO: FIXME
```
jorshi, line 365: Is this still broken?

heareval/predictions/task_predictions.py
jorshi: I think I get it -- see comment below

jorshi: I can also see how this is awkward for event based metrics :/

jorshi (approved):
This overall looks good -- made a few comments / questions. I think the score name for the event detection metrics needs to be addressed and update the docstring on the score module before merge but looks good.

heareval/score.py
turian: dcase is currently way broken in this PR in a variety of ways, I am fixing it in upcoming PRs

heareval/evaluation/task_evaluation.py
turian: AUC doesn't really make sense for multiclass in the way we thought, unless we want macroaveraging

turian: Removed

heareval/evaluation/task_evaluation.py
turian: Sure will do. We might end up removing this file or not having this as a separate step from prediction.

heareval/predictions/task_predictions.py
turian: Yeah. In a later PR I am accumulating at the end of validation, not every validation step. But it's still a pain for event-based metrics.

heareval/predictions/task_predictions.py
turian: Yeah so you can choose the primary score for early stopping and checkpointing. But you'll see in an upcoming PR that I will accumulate the validation score over the entire epoch.

heareval/predictions/task_predictions.py
turian: yes. we still aren't giving the user any option to specify what device they want for training. we're still missing model selection, etc.

PR #170, opened 2021-08-14
Add Validation and Test Split in pipeline

khumairraj:


heareval/tasks/dcase2016_task2.py
```diff
@@ -88,12 +96,19 @@ def get_split_metadata(self, split: str) -> pd.DataFrame:
                 .replace("annotation", "sound")
                 .replace(".txt", ".wav")
             )
-            assert os.path.exists(sound_file)
```
khumairraj, line 91: This has been removed here but is being addressed in #162

heareval/tasks/dcase2016_task2.py
```diff
@@ -9,17 +9,21 @@
...
 import pandas as pd
 
 import heareval.tasks.pipeline as pipeline
+import heareval.tasks.util.luigi as luigi_utils
 
 logger = logging.getLogger("luigi-interface")
 
+# This percentage should not be change as this decides
+# the data in the split and hence is not a part of the config
+VALIDATION_PERCENTAGE = 0.3
```
turian, line 25: 0.2

heareval/tasks/dcase2016_task2.py
```diff
@@ -67,6 +72,9 @@ def requires(self):
     }
 
     def get_split_metadata(self, split: str) -> pd.DataFrame:
+        # Since the valid is part of the train
```
turian, line 75: I don't understand what this does or why this works can you explain better? It's very confusing

heareval/tasks/dcase2016_task2.py
```diff
@@ -9,17 +9,21 @@
...
 
 import luigi
 import pandas as pd
 
 import heareval.tasks.pipeline as pipeline
+import heareval.tasks.util.luigi as luigi_utils
 
 logger = logging.getLogger("luigi-interface")
 
+# This percentage should not be change as this decides
```
turian, line 23: I mean shouldn't this just be global and not task specific? If there is not validation, use 20% as validation. If there is not test, use 20% as test. I don't see why it should be task specific

heareval/tasks/dcase2016_task2.py
```diff
@@ -88,12 +96,19 @@ def get_split_metadata(self, split: str) -> pd.DataFrame:
...
                 .replace(".txt", ".wav")
             )
-            assert os.path.exists(sound_file)
             metadata = metadata.assign(
                 relpath=sound_file,
                 slug=lambda df: df.relpath.apply(self.slugify_file_name),
-                split=lambda df: split,
                 subsample_key=lambda df: self.get_subsample_key(df),
+                # For train the split is decided by the which set function.
+                # which takes in validation and testing percentage
```
turian, line 104: Cool

heareval/tasks/dcase2016_task2.py
khumairraj: Actually the get split metadata was using the split to iterate over and generate. But since there is no requires task for the valid split, this was supposed to be sampled. It is clear in the new commits

heareval/tasks/dcase2016_task2.py
khumairraj: I have moved it into the pipeline and made a generic split function that will handle all splits scenarios with the global percentage for test and validation. This variable can be moved out of the `pipeline.py` in a better place, but logically is a global variable now.

turian:
@khumairraj this is awesome. Did you try running it and making sure from the JSON statistics outputs that it looks right?

turian (approved):

PR #176, opened 2021-08-15
Fixes to get DCASE evaluation working

turian:
`task_predict.py` is really long and mixes a lot of different logic together. I'm wondering the most logical way to break it down. Perhaps to move `get_events_for_all_files` and its child to another file? I'm not sure honestly.

### DIFF
--- a/heareval/predictions/task_predictions.py
+++ b/heareval/predictions/task_predictions.py
@@ -97,7 +97,7 @@ class PredictionModel(pl.LightningModule):
     def training_step(self, batch, batch_idx):
         # training_step defined the train loop.
         # It is independent of forward
-        x, y, filename, timestamp = batch
+        x, y, _ = batch
         y_hat = self.predictor.forward_logit(x)
         # Why do we need .float()?
         # https://discuss.pytorch.org/t/multi-label-binary-classification-result-type-float-cant-be-cast-to-the-desired-output-type-long/117915
@@ -107,16 +107,16 @@ class PredictionModel(pl.LightningModule):
         return loss
 
     def validation_step(self, batch, batch_idx):
-        x, y, filenames, timestamps = batch
+        x, y, metadata = batch
         y_hat = self.predictor.forward_logit(x)
         y_pr = self.predictor(x)
-        return {
+        z = {
             "predictions": y_pr,
             "predictions_logit": y_hat,
             "targets": y,
-            "filenames": filenames,
-            "timestamps": timestamps,
         }
+        # https://stackoverflow.com/questions/38987/how-do-i-merge-two-dictionaries-in-a-single-expression-taking-union-of-dictiona
+        return {**z, **metadata}
 
     def validation_epoch_end(self, outputs):
         targets = []
@@ -128,8 +128,8 @@ class PredictionModel(pl.LightningModule):
             targets += output["targets"]
             predictions += output["predictions"]
             predictions_logit += output["predictions_logit"]
-            filenames += output["filenames"]
-            timestamps += output["timestamps"]
+            filenames += output["filename"]
+            timestamps += output["timestamp"]
 
         targets = torch.stack(targets)
         predictions = torch.stack(predictions)
@@ -246,8 +246,7 @@ class SplitMemmapDataset(Dataset):
             return (
                 x,
                 torch.zeros((self.nlabels,), dtype=torch.int32),
-                filename,
-                timestamp,
+                {"filename": filename, "timestamp": timestamp},
             )
         # TODO: Could rewrite faster using scatter_:
         # https://discuss.pytorch.org/t/what-kind-of-loss-is-better-to-use-in-multilabel-classification/32203/4
@@ -258,8 +257,7 @@ class SplitMemmapDataset(Dataset):
             .values
             # BCEWithLogitsLoss wants float not long targets
             .float(),
-            filename,
-            timestamp,
+            {"filename": filename, "timestamp": timestamp},
         )
 
 
--- a/heareval/predictions/task_predictions.py
+++ b/heareval/predictions/task_predictions.py
@@ -99,9 +99,7 @@ class PredictionModel(pl.LightningModule):
         # It is independent of forward
         x, y, _ = batch
         y_hat = self.predictor.forward_logit(x)
-        # Why do we need .float()?
-        # https://discuss.pytorch.org/t/multi-label-binary-classification-result-type-float-cant-be-cast-to-the-desired-output-type-long/117915
-        loss = self.predictor.logit_loss(y_hat, y.float())
+        loss = self.predictor.logit_loss(y_hat, y)
         # Logging to TensorBoard by default
         self.log("train_loss", loss)
         return loss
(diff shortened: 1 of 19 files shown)


heareval/predictions/task_predictions.py
```diff
@@ -91,22 +97,74 @@ def forward(self, x):
...
-        self.log("val_loss", loss, prog_bar=True)
+        return {
+            "predictions": y_pr,
+            "predictions_logit": y_hat,
+            "targets": y,
+            "filenames": filenames,
+            "timestamps": timestamps,
+        }
+
+    def validation_epoch_end(self, outputs):
```
jorshi, line 121: Cool, good idea to add the validation step here. Instead of trying to fit this all into a single predictor class -- let's have a EventPredictionModel and a ScenePredictionModel?

heareval/predictions/task_predictions.py
```diff
@@ -215,9 +307,33 @@ def create_events_from_prediction(
...
     predictions = (predictions > threshold).astype(np.int8)
+    # print("predictions", predictions)
 
+    event_tree = IntervalTree()
+    for label in range(predictions.shape[1]):
+        # print(label, predictions[:, label])
+        # print(np.where(predictions[:, label]))
+        for group in more_itertools.consecutive_groups(
+            np.where(predictions[:, label])[0]
+        ):
```
jorshi, line 318: Nice -- this looks clean, wasn't sure how consecutive groups worked -- so this looks for consecutive groups of ones?

heareval/predictions/task_predictions.py
```diff
@@ -146,32 +204,62 @@ def __init__(
         self.labels = pickle.load(
             open(embedding_path.joinpath(f"{split_name}.target-labels.pkl"), "rb")
         )
+        # Only used for event-based prediction
```
jorshi, line 207: Same here -- different classes of DatasetLoaders for either event or prediction types?

jorshi: Then we can avoid all the logic in these functions that is looking for either event or scene.

heareval/predictions/task_predictions.py
turian: yes

heareval/predictions/task_predictions.py
turian: Done

heareval/predictions/task_predictions.py
turian: I think I was able to make this clean by having a dataset metadata dict. It was very simple to change the code to use this. The only downside really is a bit less type-safety and things are a little less explicit.

heareval/predictions/task_predictions.py
```diff
@@ -91,28 +97,160 @@ def forward(self, x):
...
+        z = {
+            "prediction": y_pr,
+            "prediction_logit": y_hat,
+            "target": y,
+        }
+        # https://stackoverflow.com/questions/38987/how-do-i-merge-two-dictionaries-in-a-single-expression-taking-union-of-dictiona
+        return {**z, **metadata}
+
+    # Implement this for each inheriting class
+    # TODO: Can we combine the boilerplate for both of these?
```
turian, line 121: I'm not sure if we can simplify the boilerplate in validation_epoch_end implementations further

heareval/predictions/task_predictions.py
```diff
@@ -146,38 +286,64 @@ def __init__(
...
             torch.nn.functional.one_hot(torch.LongTensor(y), num_classes=self.nlabels)
             .max(axis=0)
-            .values,
+            .values.float(),
+            self.metadata[idx],
         )
 
 
 def create_events_from_prediction(
     prediction_dict: Dict[float, torch.Tensor],
```
turian, line 341: Yeah so this params shouldn't be hardcoded in predict really, it should be a part of the event scoring classes. I am not really sure how to move these hyperparams to make them in scores.py but that's where they belong.

Also these methods are really kind confusing here, I think they belong somewhere else.

heareval/predictions/task_predictions.py
```diff
@@ -15,10 +15,13 @@
 """
```
turian, line 15: This file is too long, but I don't want to break it into ten files just for the sake of it. I wonder what is the most logical way to break it down.

heareval/predictions/task_predictions.py
```diff
@@ -216,44 +386,32 @@ def create_events_from_prediction(
...
+
+            start = timestamps[startidx]
+            end = timestamps[endidx]
+            # Add event if greater than the minimum duration threshold
+            if end - start >= min_duration:
+                # We probably don't need this interval tree and can remove it maybe
+                # from requirements
+                event_tree.addi(begin=start, end=end, data=label)
+
+    # Er why return an intervaltree when we immediately postprocess it to a dict?
```
turian, line 404: I still want to remove this intervaltree nonsense, it doesn't add anything as far as I can tell?
I don't think we need it at all

heareval/predictions/task_predictions.py
```diff
@@ -349,31 +512,62 @@ def task_predictions_train(
...
 
-    checkpoint_callback = ModelCheckpoint(monitor=target_score, mode="max")
+    if scores[0].maximize:
+        mode = "max"
+    else:
+        mode = "min"
+    checkpoint_callback = ModelCheckpoint(monitor=target_score, mode=mode)
     early_stop_callback = EarlyStopping(
-        monitor=target_score, min_delta=0.00, patience=10, verbose=False, mode="max"
+        # TODO: Tune these
```
turian, line 547: Still TODO In upcoming PR

heareval/predictions/task_predictions.py
```diff
@@ -349,31 +512,62 @@ def task_predictions_train(
...
     )
     valid_dataloader = dataloader_from_split_name(
-        "valid", embedding_path, label_to_idx, nlabels
+        "valid", embedding_path, label_to_idx, nlabels, metadata["embedding_type"]
     )
     trainer.fit(predictor, train_dataloader, valid_dataloader)
     return predictor
 
 
+# This all needs to be cleaned up and simplified later.
```
turian, line 568: Still todo in upcoming PR

heareval/score.py
```diff
@@ -171,19 +191,34 @@ class EventBasedScore(SoundEventScore):
...
 
 available_scores: Dict[str, Callable] = {
-    "top1_err": Top1Error,
-    "pitch_err": partial(Top1Error, name="pitch_err"),
-    "chroma_err": ChromaError,
-    "onset_only_event_based": partial(
-        EventBasedScore, params={"evaluate_offset": False, "t_collar": 0.2}
+    "top1_acc": Top1Accuracy,
+    "pitch_acc": partial(Top1Accuracy, name="pitch_acc"),
+    "chroma_acc": ChromaAccuracy,
```
turian, line 204: The event glomming hyperparams belong here too somehow.

The score names are getting pretty long but at least they're specific

jorshi: Ya cool -- this is sort of how I imagined this would go, I had thought about moving these score params into the task config dict where we specify the eval metric -- but I think this is okay here.

heareval/predictions/task_predictions.py
jorshi: Could we move these into the score function somehow? Like a sed eval function that receives a set of frame embedding predictions / targets and then creates the events from those and calculates the scores? So we don't need to fuss with the conversion beforehand?

heareval/predictions/task_predictions.py
jorshi: One of the remaining things I think we get from intervaltree is the sort of all events by the start timestamp -- so that is kind of nice. I'm curious whether that is necassary for the sed eval stuff. Either way, that wouldn't be too challenging to implement if it is requirement and we have everything we need it looks like to go straight from the consecutive groups to a dictionary w/o the intervaltree

jorshi: So if you wanna ditch the interval tree I'd say go for it

heareval/predictions/task_predictions.py
jorshi: Ya it's a bit long -- but not the worst. I was going to say we could move the prediction models to there own module -- but maybe it is kind of nice to have them in the same file. I don't think it is a huge issue. I'm kind of into the idea of moving the event creation from frames functions into the scoring modules.

heareval/predictions/task_predictions.py
jorshi: hmm ya. interesting -- on one hand it creates some flexible down the road if we want to pass other types of metadata depending on the task type. It's not ideal have to pass around the empty dict for the scene stuff. I can't think of a better approach atm though, this seems to be working pretty well.

jorshi (commented):
Took an initial pass on this -- overall awesome work, I think this is looking good. Made a few comments.

heareval/predictions/task_predictions.py
turian: The previous implementation had a bug on the boundaries

heareval/predictions/task_predictions.py
khumairraj: Yes. The metadata dict looks nice and it does give more flexibility.

heareval/predictions/task_predictions.py
khumairraj: We might want to use this to dump the events in the test part (I see commented code for the test). But putting in score also sounds like a good option

khumairraj (approved):
I think the abstraction are great and the event-based task are fitting in well.

PR #184, opened 2021-08-17
Improved README and install

turian:


heareval/tasks/pipeline.py
```diff
@@ -801,18 +803,23 @@ def run(self):
...
-        shutil.copytree(
             self.requires()["vocabmeta"].workdir,
-            self.workdir,
-            dirs_exist_ok=True,
-        )
+        ]:
+            dst = self.workdir
+            for item in os.listdir(src):
+                if item.endswith(".csv"):
+                    continue
```
khumairraj, line 814: We have to avoid copying the .csv only in the traintestmeta as we are using the json versions from this folder. For vocabmeta folder we need the .csv files as well.

khumairraj (approved):

heareval/tasks/speech_commands.py
```diff
@@ -104,13 +104,15 @@ def run(self):
         for file_obj in train_path.iterdir():
             if file_obj.is_dir() and file_obj.name != BACKGROUND_NOISE:
                 linked_folder = Path(os.path.join(self.workdir, file_obj.name))
-                linked_folder.unlink(missing_ok=True)
+                if linked_folder.exists():
+                    linked_folder.unlink(missing_ok)
```
khumairraj, line 108: `missing_ok` is not defined here. Did you intend to remove it?

heareval/tasks/pipeline.py
khumairraj: Fixed this.

heareval/tasks/speech_commands.py
turian: oops yes I should have removed it

PR #187, opened 2021-08-17
[WIP] Adding tests for the pipeline

khumairraj:


turian:
Yeah I'm confused by this. Like if you're adding more unit tests, great.

But the idea is that we should have a github action which runs all the key scripts in heareval to preprocess and then do evaluation, and should throw an exception if there's an error.

So really you just need like a __main__ function that runs everything and to call that. Forget about pytest

PR #189, opened 2021-08-18
[WIP] Integrations Test and changing run end to end

khumairraj:


heareval/embeddings/runner.py
```diff
@@ -18,16 +19,33 @@
...
     "--tasks-dir",
     default="tasks",
     help="Location of tasks to compute embeddings on",
     type=str,
 )
-def runner(module: str, model: str = None, tasks_dir: str = "tasks") -> None:
+@click.option(
+    "--task-version",
+    default="all",
+    help="Versioned task name in the tasks dir to compute embedding on",
```
turian, line 32: This option makes no sense to me. Why are we versioning the task name?

heareval/embeddings/runner.py
```diff
@@ -18,16 +19,33 @@
     help="Location of model weights file",
     type=click.Path(exists=True),
 )
+@click.option("--version", default="", help="Version of module and weights")
```
turian, line 22: I don't get this either. Where is this spec'ed

heareval/embeddings/runner.py
```diff
@@ -36,12 +54,16 @@ def runner(module: str, model: str = None, tasks_dir: str = "tasks") -> None:
         )
 
     # Load the embedding model
-    embedding = Embedding(module, model)
+    embedding = Embedding(module, model, str(version))
```
turian, line 57: Is this documented in the API anywhere? I mean maybe this makes sense but where is the API documentation? Otherwise I'm not sure we can just slip this change in without announcing it or whatever

heareval/embeddings/task_embeddings.py
```diff
@@ -81,7 +82,7 @@ def name(self):
         # TODO: would be nice to include version in this string, a versioned string.
         #   Potentially can set a version from the command line too to help with testing
         #   the same model but with difference versions of the weights.
```
khumairraj, line 84: We have a todo here which specifies the task version.

heareval/embeddings/runner.py
khumairraj: This is specified in the todo at two places. Actually, this is just for convenience. For example, someone has one version of the model and produces embedding for that. And then retrains the model and has another set of weights. This will help them have versions for that. Basically, in the current flow -> Product embeddings which are for a model version, specify the version in the command call, otherwise there will no be no version as it was previously. That is the only thing this does.

turian:
https://github.com/neuralaudio/hear-eval-kit/pull/200

Issue #193, opened 2021-08-22
Subsample tasks based on number of hours


jorshi:
Currently we can subsample all splits using the dataset_fraction attribute in task config, but we want to select the exact number of hours each split should contain. 

- Remove dataset_fraction 
- Add a global task config to set the max task duration in seconds (i.e. 5hrs)
- Train/test/val split percentage should also be a global variable 

This leads to a max number of audio files per split, should be rounded to the closest number of audio files.

PR #195, opened 2021-08-23
Max task duration

turian:
* Document how each of the steps of the preprocess pipeline works.
* MAX_TASK_DURATION_PER_SPLIT is 5 hours of audio for the entire dataset. dataset_fraction is gone.
* A handful of the tasks are renamed and clarified

Closes https://github.com/neuralaudio/hear-eval-kit/issues/193

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,8 @@ Each pipeline will download and preprocess each dataset according
 to the following DAG:
 * DownloadCorpus
 * ExtractArchive
-* ExtractMetadata
+* ExtractMetadata: Create splits over the entire corpus and find
+the label metadata for them.
 * SubsampleSplit (subsample each split) => MonoWavTrimCorpus => SplitData (symlinks)
 * SplitData => {SplitMetadata, ResampleSubcorpus}
 * SplitMetadata => MetadataVocabulary
--- a/README.md
+++ b/README.md
@@ -121,6 +121,15 @@ python3 -m heareval.tasks.runner all
 You can also just run individual tasks:
 python3 -m heareval.tasks.runner [speech_commands|nsynth_pitch|dcase2016_task2]
 
+Each pipeline will download and preprocess each dataset according
+to the following DAG:
+* download and extract datafiles
+* ExtractMetadata
+* SubsampleSplit (subsample each split) => MonoWavTrimCorpus => SplitData (symlinks)
+* SplitData => {SplitMetadata, ResampleSubcorpus}
+* SplitMetadata => MetadataVocabulary
+* FinalizeCorpus
+
 These commands will download and preprocess the entire dataset. An
 intermediary directory defined by the option `luigi-dir`(default
 `_workdir`) will be created, and then a final directory defined by
@@ -183,6 +192,7 @@ deterministically with the following command:
 python3 -m heareval.tasks.sampler <taskname>
 ```
 
+[rewrite]
 **_NOTE_** : Each task config has `dataset_fraction`. The data in
 each split is subsampled by this fraction in the final output. This
 is not to be confused with the `--small` flag which is used to run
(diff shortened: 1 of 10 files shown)


jorshi (commented):
I'm good with all of these changes and renames -- especially the simplifications. 

I just ran this fully on my computer. Subsampling for DCASE isn't working -- ending up with an incorrect number of files in each. I believe this is caused by the stratified subsampling. There is a stratify_key for dcase which is based on the label and I'm not sure if that is behaving how we expect it to when it gets to this function: https://github.com/neuralaudio/hear-eval-kit/blob/15e7710dac72c49b1504880bb0765d468a0393a0/heareval/tasks/util/luigi.py#L201 (called from the run method in SubsampleSplit)-- I didn't dig into it too much yet, but I think something weird is going on there when it comes to dcase. Not sure how you would properly subsample the event-based files to make sure that all the labels are being represented (this won't be an issue for the train / valid because combined they are going to be much less than 4hrs).

Also, I think because of the stratified subsampling we're ending up with less than 5hrs of audio for nsynth as well.

khumairraj (approved):
The renaming of the variables looks good.

khumairraj:
> I'm good with all of these changes and renames -- especially the simplifications.
> 
> I just ran this fully on my computer. Subsampling for DCASE isn't working -- ending up with an incorrect number of files in each. I believe this is caused by the stratified subsampling. There is a stratify_key for dcase which is based on the label and I'm not sure if that is behaving how we expect it to when it gets to this function:
> 
> https://github.com/neuralaudio/hear-eval-kit/blob/15e7710dac72c49b1504880bb0765d468a0393a0/heareval/tasks/util/luigi.py#L201
> 
> (called from the run method in SubsampleSplit)-- I didn't dig into it too much yet, but I think something weird is going on there when it comes to dcase. Not sure how you would properly subsample the event-based files to make sure that all the labels are being represented (this won't be an issue for the train / valid because combined they are going to be much less than 4hrs).
> Also, I think because of the stratified subsampling we're ending up with less than 5hrs of audio for nsynth as well.

I think I get the problem is --
```
    grp_count = metadata["stratify_key"].value_counts() * max_files / len(metadata)
```
In the above line, for each stratify key, the maximum number of files are found. The assumption is that each row is associated with one file. However, for event based task, this is not true.

jorshi:
> I'm good with all of these changes and renames -- especially the simplifications.
> 
> I just ran this fully on my computer. Subsampling for DCASE isn't working -- ending up with an incorrect number of files in each. I believe this is caused by the stratified subsampling. There is a stratify_key for dcase which is based on the label and I'm not sure if that is behaving how we expect it to when it gets to this function:
> 
> https://github.com/neuralaudio/hear-eval-kit/blob/15e7710dac72c49b1504880bb0765d468a0393a0/heareval/tasks/util/luigi.py#L201
> 
> (called from the run method in SubsampleSplit)-- I didn't dig into it too much yet, but I think something weird is going on there when it comes to dcase. Not sure how you would properly subsample the event-based files to make sure that all the labels are being represented (this won't be an issue for the train / valid because combined they are going to be much less than 4hrs).
> Also, I think because of the stratified subsampling we're ending up with less than 5hrs of audio for nsynth as well.

To give a bit more detail on the issue that I am seeing. After running this, these are the file counts that I am getting:
Speech Commands (1s audio clips):
- test: 3593 
- train: 10794
- valid: 3594

Nsynth Pitch (4s audio clips):
- test: 849 (3396s)
- train: 2654 (10616s)
- valid: 856 (3424s)

DCASE (120s audio clips):
- test: 27 (3240s)
- train: 9 (1080s)
- valid: 3 (360s)

DCASE is the most concerning because subsampling should not be occurring at all for the train and valid sets. The dev set does not contain over 4 hrs of audio.

khumairraj:
#204 -> Remove dcase stratified subsampling.
The file counts now are - 
DCASE (120s audio clips):
test: 30 (3600)
train: 13 (1560)
valid: 5 (600)

#205 -> Fixes the slug bug.

PR #204, opened 2021-08-23
Remove stratification across pipeline

khumairraj:


heareval/tasks/dcase2016_task2.py
```diff
@@ -106,7 +106,8 @@ def get_split_metadata(self, split: str) -> pd.DataFrame:
                 subsample_key=lambda df: self.get_subsample_key(df),
                 split=lambda df: split,
                 split_key=lambda df: self.get_split_key(df),
-                stratify_key=lambda df: self.get_stratify_key(df),
+                # No stratified sampling for event based task
+                stratify_key="none",
```
turian, line 110: Okay so I don't love this either. Like why disable stratification for just DCASE or just event tasks? Why not disable stratification when there are few examples? Or better yet, why not just disable stratification for EVERYTHING?

My preference would be that we disable stratification for EVERYTHING, in ONE generic place, not task by task. But have a comment explaining what disabling does, and what happens if we DON'T disable.

turian: With more detailed text on what stratitifcation does and doesn't do

heareval/tasks/pipeline.py
```diff
@@ -171,7 +171,9 @@ def slugify_file_name(relative_path: str):
 
         The basic version here takes the filename and slugifies it.
         """
-        return f"{slugify(str(Path(relative_path).stem))}"
+        slug_text = str(Path(relative_path).stem)
+        slug_text = slug_text.replace("-", "_negative_")
+        return f"{slugify(slug_text)}"
 
     @staticmethod
     def get_stratify_key(df: DataFrame) -> Series:
```
turian, line 179: Where is stratify_key even used? I was searching the code and can't find it :|

jorshi: https://github.com/neuralaudio/hear-eval-kit/blob/4de804e091dab4c2e05385044b44aa9737b66d4c/heareval/tasks/util/luigi.py#L215

heareval/tasks/dcase2016_task2.py
jorshi: Ya can we just pull out the stratified subsampling for now? It's a good idea, but needs to be thought out more thoroughly for non-multiclass datasets. I think a stable random subsampling of max_files in SubsampleSplit is great for now.

jorshi: Then we will ensure that we have as close to 5hrs of data.

khumairraj: Yes. @turian, the problem is not with fewer data in the event example, it is the multilabel setup of event-based data that stratification cannot work(even if there were lots of data points). Maybe we can disable the stratification across task if we want to. Normal stratification will work for single-label data only.
1. Either we disable the stratification for not-single label tasks(multilabel and event-based)
2. We remove stratification for all tasks for now.

heareval/tasks/pipeline.py
khumairraj: Yes. In the above line, passing `none` in the stratify key, doesn't stratify and treats the key as not useful.

heareval/tasks/dcase2016_task2.py
turian: @khumairraj let's disable stratification across all tasks for now. Let's comment out the stratification code and leave a detailed comment what it does and what the limitations are, and why it is disabled.

heareval/tasks/pipeline.py
turian: That also needs a comment explaining

heareval/tasks/dcase2016_task2.py
khumairraj: Removed the stratification from the pipeline and made comments on the function and limitation

turian (approved):

Issue #203, opened 2021-08-23
Bug in slugifying filenames with dashes


jorshi:
In the luigi pipeline metadata dataframe we slugify the relative path of the filename. This is broken for filenames that contain a dash character. For example: `test_1_ebr_-6_nec_4_poly_1.wav` in dcase. This slugifies to `test-1-ebr-6-nec-4-poly-1` , which is the same as what  `test_1_ebr_6_nec_4_poly_1.wav` slugifies to.

jorshi:
Potential solution is to use the replacement arg in slugify (we would override the `slugify_file_name` method in ExtractMetadata for the dcase task) to replace `-` with `negative_`. See https://github.com/un33k/python-slugify

![dcase-slugs-potential-solution](https://user-images.githubusercontent.com/4112059/130492356-93f98430-495e-4cef-bbb0-0b2ef86fb8ab.png)

jorshi:
We should also do a sanity check in the ExtractMetadata run function to make sure that all the slugs are unique. i.e. something like:

```
assert len(process_metadata["relpath"].unique()) == len(process_metadata["slug"].unique())
```

PR #205, opened 2021-08-23
Fix Unique Slug

khumairraj:
#203

### DIFF
--- a/heareval/tasks/pipeline.py
+++ b/heareval/tasks/pipeline.py
@@ -293,12 +293,26 @@ class ExtractMetadata(WorkTask):
     def run(self):
         # Process metadata gets the final metadata to be used for the task
         process_metadata = self.get_process_metadata()
+
         # Check if one slug is associated with only one relpath.
         # Also implies there is a one to one correspondence between relpath and slug.
+        #  1. One slug to one relpath -- the bug which we were having is one slug for
+        #   two relpath(relpath with -6 as well as +6 having the same slug), groupby
+        #   by slug and see if one relpath is associated with one slug - this is done
+        #   in the assert statement.
+        #  2. One relpath to one slug -- always the case, because slugify is
+        #   a deterministic function.
+        #  3. relpath.nunique() == slug.nunique(), automatically holds if the above
+        #   two holds.
         assert (
             process_metadata.groupby("slug")["relpath"].nunique() == 1
         ).all(), "One slug is associated with more than one file"
         "Please make sure slugs are unique at a file level"
+
+        # Assertion sanity check -- one to one mapping between the relpaths and slugs
+        assert (
+            process_metadata["relpath"].nunique() == process_metadata["slug"].nunique()
+        )
 
         # Split the metadata to create valid and test set from train if they are not
         # created explicitly in the get process metadata
--- a/heareval/tasks/pipeline.py
+++ b/heareval/tasks/pipeline.py
@@ -291,11 +291,14 @@ class ExtractMetadata(WorkTask):
         return metadata
 
     def run(self):
+        # Process metadata gets the final metadata to be used for the task
         process_metadata = self.get_process_metadata()
-        # Check if one slug is associated with only one relpath
+        # Check if one slug is associated with only one relpath.
+        # Also implies there is a one to one correspondence between relpath and slug.
         assert (
             process_metadata.groupby("slug")["relpath"].nunique() == 1
-        ).all(), "There are duplicate slugs for one audio file"
+        ).all(), "One slug is associated with more than one file"
+        "Please make sure slugs are unique at a file level"
 
         # Split the metadata to create valid and test set from train if they are not
         # created explicitly in the get process metadata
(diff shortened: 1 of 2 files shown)


heareval/tasks/pipeline.py
```diff
@@ -171,7 +171,9 @@ def slugify_file_name(relative_path: str):
 
         The basic version here takes the filename and slugifies it.
         """
-        return f"{slugify(str(Path(relative_path).stem))}"
+        slug_text = str(Path(relative_path).stem)
+        slug_text = slug_text.replace("-", "_negative_")
```
turian, line 175: Add a comment to explain why

heareval/tasks/pipeline.py
```diff
@@ -171,7 +171,9 @@ def slugify_file_name(relative_path: str):
...
+        return f"{slugify(slug_text)}"
```
turian, line 176: speech_commands overwrites slugify_file_name. That is really risky for a bug.

Either:
1) speech commands does not override and it uses the same
or
2) the generic class method has slugify and is a *wrapper* method that fixes the - in the  relpath and calls a second method that does slugify on the wrapper fixed relpath, which speech commands also uses.

heareval/tasks/pipeline.py
```diff
@@ -291,6 +293,11 @@ def split_train_test_val(self, metadata: pd.DataFrame):
 
     def run(self):
         process_metadata = self.get_process_metadata()
```
turian, line 295: There are a handful of things I hate about this :(

ExtractMetadata.get_process_metadata() is a generic method but:
* there is no comment describing WHAT this method should do, and when it should be overriden
* the generic method doesn't use slugify_file_name. Why?

This seems confusing and risky and bug prone

heareval/tasks/pipeline.py
```diff
@@ -291,6 +293,11 @@ def split_train_test_val(self, metadata: pd.DataFrame):
...
+        # Check if one slug is associated with only one relpath
+        assert (
+            process_metadata.groupby("slug")["relpath"].nunique() == 1
+        ).all(), "There are duplicate slugs for one audio file"
+
```
turian, line 300: I think this confusing that this isn't where slugify_file_name is called and I worry that overriding methods might skip this check or do it wrong.

Also I think this assert is wrong. We don't want different slugs for one relpath to have nunique 1.

We want relpath.nunique == slug.nunique for ALL relpaths.

And we want that if we map from all replath to all slug, that there is a one to one mapping between slug and relpath

heareval/tasks/pipeline.py
jorshi: This should not be here. dcase should overwrite slugify_file_name like speech_commands does.

heareval/tasks/pipeline.py
jorshi: dcase is a special case here using "-" to refer to a negative, this method should stay the same and dcase should override slugify_file_name with the changes above.

heareval/tasks/pipeline.py
jorshi: i.e. this case of replacing "-" with "_negative_" is specific only to dcase

heareval/tasks/pipeline.py
jorshi: I agree -- this needs to be documented better - and probably simplified. But let's save the refactor for a different PR.

heareval/tasks/pipeline.py
jorshi: run is always called and is what dispatches the rest of the pipelines. slugify_file_name is a helper that can be used in deriving classes. So this is the correct place for the check.

We definitely want relpath.nunique == slug.nunique.

khumairraj: Yes. `run()` is never overridden, so the test can be here.
For the assert, cases - 
1. `One slug for one relpath` -- the bug which we were having is one slug for two relpath(relpath with -6 as well as +6 having the same slug), groupby by slug and see if one relpath is associated with one slug - this is done in the `assert` statement.
2. `One relpath for one slug` -- this is always true, as the slug function is deterministic.
3. `relpath.nunique() == slug.nunique()`, automatically holds if the above two holds.

So, by the above reasoning, the current assertion is sufficient. But please let me know if I miss some case.

heareval/tasks/pipeline.py
khumairraj:
```
    def get_process_metadata(self) -> pd.DataFrame:
        """
        Return a dataframe containing the task metadata for this
        entire task.
        By default, we do one split at a time and then concat them.
        This runs only for the splits which have a requires task. All
        other splits are sampled with the split train test val function.
        You might consider overriding this for some datasets (like
        Google Speech Commands) where you cannot process metadata
        on a per-split basis.
        """
```
The above description gives the information about this method, and also when to override it. But we can be more specific.

heareval/tasks/pipeline.py
khumairraj: Yes. We can override for `dcase` slugify to make it more explicit, and explainable to why we are doing this. Because other tasks(in general) might have `-` in their name, and we might not want to convert that to `-`. Thanks for the suggestion. Will change

heareval/tasks/pipeline.py
turian: @jorshi and @khumairraj yes okay let's postpone refactor: https://github.com/neuralaudio/hear-eval-kit/issues/208

jorshi (approved):

PR #225, opened 2021-08-26
[WIP] Test score sanity check

turian:
https://github.com/neuralaudio/hear-eval-kit/pull/223 implements test scoring in the predictor runner.

We also have evaluation runner, which takes targets and prediction pickle files and scores them.

This is a good sanity check.

However, we are getting different scores in predictor than in runner. Perhaps the prediction is pickling the targets in the wrong order or something? Help!


turian:
https://github.com/neuralaudio/hear-eval-kit/pull/243

PR #226, opened 2021-08-26
Grid search for predictor

turian:
Also enable GPU, and wandb logging.


khumairraj (approved):
That looks great! The random selection and its clean setup.

PR #243, opened 2021-08-28
Prediction test grid pickle

turian:
#223 implements test scoring in the predictor runner.

We also have evaluation runner, which takes targets and prediction pickle files and scores them.

This is a good sanity check.

### DIFF
--- a/heareval/predictions/task_predictions.py
+++ b/heareval/predictions/task_predictions.py
@@ -33,7 +33,12 @@ from sklearn.model_selection import ParameterGrid
 from torch.utils.data import DataLoader, Dataset
 from tqdm.auto import tqdm
 
-from heareval.score import ScoreFunction, available_scores, label_vocab_as_dict
+from heareval.score import (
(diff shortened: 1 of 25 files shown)


heareval/predictions/task_predictions.py
```diff
@@ -25,15 +25,20 @@
 import pandas as pd
 import pytorch_lightning as pl
 import torch
+import wandb
```
jorshi, line 28: What does wandb do? Have we added to the setup.py requirements?

jorshi: I see it was already imported below so I guess this was added in a previous PR :)

jorshi (approved):

heareval/predictions/task_predictions.py
turian: @jorshi better than tensorboard and automatically logs a lot of performance metrics like disk writes, GPU memory transfer, etc.

PR #270, opened 2021-08-30
Much bigger gridsearch

turian:
We won't use this for the final leaderboard. Instead, we'll make a much tighter grid by running these on strong benchmarks.

### DIFF
--- a/heareval/embeddings/task_embeddings.py
+++ b/heareval/embeddings/task_embeddings.py
@@ -38,7 +38,7 @@ from intervaltree import IntervalTree
 from torch.utils.data import DataLoader, Dataset
 from tqdm.auto import tqdm
 
-import wandb
+# import wandb
 from heareval.tasks.pipeline import SPLITS
 
 TORCH = "torch"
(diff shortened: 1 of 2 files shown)


heareval/predictions/task_predictions.py
```diff
@@ -742,13 +752,26 @@ def print_scores(mode, scores_and_trainers):
         scores_and_trainers.append((best_model_score, trainer, predictor))
         print_scores(mode, scores_and_trainers)
 
+    def vstr(v):
```
jorshi, line 755: vstr stands for value str? This is for serializing values to prepare for json? I know you want this short but the intentions of this function aren't super clear. `serialize_value(str: v)`?

heareval/embeddings/task_embeddings.py
```diff
@@ -38,7 +38,7 @@
 from torch.utils.data import DataLoader, Dataset
 from tqdm.auto import tqdm
 
-import wandb
+# import wandb
```
jorshi, line 41: What's up with the wandb stuff here?

jorshi (approved):

heareval/predictions/task_predictions.py
turian: Done, good idea

heareval/embeddings/task_embeddings.py
turian: I'm sick of auth'ing wand for every GCP run. But when you use it, you get stats charts like the following

image

PR #282, opened 2021-09-01
[WIP] Test finally works correctly, and event filtering

turian:
This works, it's just got a lot of debug output so I can test it on GPUs.

This code is gnarly and I wish I could simplify it :(

### DIFF
--- a/heareval/predictions/task_predictions.py
+++ b/heareval/predictions/task_predictions.py
@@ -33,6 +33,7 @@ import torchinfo
 from pytorch_lightning import seed_everything
 from pytorch_lightning.callbacks.early_stopping import EarlyStopping
 from pytorch_lightning.callbacks.model_checkpoint import ModelCheckpoint
+from scipy.ndimage import median_filter
 from sklearn.model_selection import ParameterGrid
 from torch.utils.data import DataLoader, Dataset
 from tqdm.auto import tqdm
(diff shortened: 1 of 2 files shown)


heareval/predictions/task_predictions.py
```diff
@@ -542,11 +591,31 @@ def get_events_for_all_files(
...
+    if postprocessing:
+        postprocess = postprocessing
+        event_dict[postprocess] = {}
+        for slug, timestamp_predictions in tqdm(event_files.items()):
+            event_dict[postprocess][slug] = create_events_from_prediction(
+                timestamp_predictions, idx_to_label, **dict(postprocess)
+            )
+    else:
+        for median_filter_ms in [0.0, 60.0, 150.0]:
+            for min_duration in [0.0, 60.0, 150.0, 250.0]:
```
jorshi, line 606: Can we not define these in the PARAM_GRID? Or, create a separate EVENT_PARAM_GRID that copies PARAM_GRID and updates it with these?

jorshi (approved):
Ya, don't love have the event based preprocessing parameters defined within the function. Would be better to define those at a higher level and pass them down if possible. But okay for now if it is working and you want to keep momentum. Def would like to refactor this though.

heareval/predictions/task_predictions.py
turian: Added

heareval/predictions/task_predictions.py
```diff
@@ -781,20 +866,29 @@ def print_scores(mode, scores_and_trainers):
...
 
     # Use that model to compute test scores
-    best_score, best_trainer, best_predictor = scores_and_trainers[0]
+    (
+        best_score,
+        best_model_path,
+        best_model_epoch,
+        best_trainer,
+        best_hparams,
+    ) = scores_and_trainers[0]
```
khumairraj, line 884: And so maybe we are picking only the last random `hparam` in the config where we ended in the `for loop` for the hyperparameter selection

heareval/predictions/task_predictions.py
```diff
@@ -781,20 +866,29 @@ def print_scores(mode, scores_and_trainers):
...
+        scores_and_trainers.append(
+            (best_model_score, model_path, epoch, trainer, hparams)
+        )
         print_scores(mode, scores_and_trainers)
 
     with open(embedding_path.joinpath("valid.grid.jsonl"), "wt") as w:
-        for score, _, predictor in scores_and_trainers:
-            w.write(json.dumps([score, hparams_to_json(predictor.hparams)]) + "\n")
+        for score, _, _, _, predictor in scores_and_trainers:
+            w.write(json.dumps([score, hparams_to_json(hparams)]) + "\n")
```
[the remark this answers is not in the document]
turian: @khumairraj should be fixed now

PR #295, opened 2021-09-05
Downstream CI

turian:
We have a few grids that are useful for development.

We add some CI/CD for downstream.


.github/workflows/pytest.yml
```diff
@@ -17,17 +17,12 @@ jobs:
       uses: actions/setup-python@master
       with:
         python-version: ${{ matrix.python-version }}
-    - name: apt-get
```
jorshi, line 20: Should we just delete this file?

jorshi (approved):

.github/workflows/pytest.yml
turian: dunno maybe we add tests? I'll remove it I guess

PR #310, opened 2021-09-27
[WIP] Add Training for Data with Folds; Do full training on test and valid before final testing

khumairraj:
This will

- Add methods to train and test on folds. Leave One out cross-validation strategy ( at a fold level ) will be applied. Each fold will be considered for testing, while all other folds will be used for training. This will lead to `n` sets of `data_splits`. Result for testing of each fold will be provided along with the aggregated result.
- Also, best grid point in the above case of `folds setup`, will only be found with the first `data_splits`. This will correspond to testing on the first fold(fold0), validation on second fold(fold1) and training on all other folds(fold2..). ( Sort the folds before deciding order(first or second) of the fold.
- For data with explicit `test, train and valid` splits defined,  best grid point will be found using the whole data.
- In both the above cases (data with folds or not with folds), the grid point will first be found followed by using this best grid point configuration to train on the `train+valid` set. Thereafter, the final testing scores will be computed.

### DIFF
--- a/heareval/predictions/task_predictions.py
+++ b/heareval/predictions/task_predictions.py
@@ -1068,14 +1068,14 @@ def data_splits_from_folds(folds: List[str]) -> List[Dict[str, List[str]]]:
     Total data splits will be equal to n, n being the total number of folds.
     Each fold will be tested by training on the remaining folds.
     """
-    folds = tuple(sorted(folds))
-    assert len(folds) == len(set(folds)), "Folds are not unique"
-    num_folds = len(folds)
+    sorted_folds: Tuple[str] = tuple(sorted(folds))
+    assert len(sorted_folds) == len(set(sorted_folds)), "Folds are not unique"
+    num_folds = len(sorted_folds)
     all_data_splits: List[Dict[str, List[str]]] = []
     for fold_idx in range(num_folds):
-        test_fold = folds[fold_idx]
-        valid_fold = folds[(fold_idx + 1) % num_folds]
-        train_folds = list(set(folds) - {test_fold, valid_fold})
+        test_fold = sorted_folds[fold_idx]
+        valid_fold = sorted_folds[(fold_idx + 1) % num_folds]
+        train_folds = list(set(sorted_folds) - {test_fold, valid_fold})
         all_data_splits.append(
             {
                 "train": train_folds,
--- a/heareval/predictions/task_predictions.py
+++ b/heareval/predictions/task_predictions.py
@@ -1068,7 +1068,7 @@ def data_splits_from_folds(folds: List[str]) -> List[Dict[str, List[str]]]:
     Total data splits will be equal to n, n being the total number of folds.
     Each fold will be tested by training on the remaining folds.
     """
-    sorted_folds: Tuple[str] = tuple(sorted(folds))
+    sorted_folds = tuple(sorted(folds))
     assert len(sorted_folds) == len(set(sorted_folds)), "Folds are not unique"
     num_folds = len(sorted_folds)
     all_data_splits: List[Dict[str, List[str]]] = []
--- a/heareval/predictions/task_predictions.py
+++ b/heareval/predictions/task_predictions.py
@@ -360,8 +360,8 @@ class EventPredictionModel(AbstractPredictionModel):
         nlabels: int,
         prediction_type: str,
         scores: List[ScoreFunction],
-        validation_target_events: Dict[str, List[Dict[str, Any]]],
-        test_target_events: Dict[str, List[Dict[str, Any]]],
+        validation_target_events: Union[Dict[str, List[Dict[str, Any]]], None],
+        test_target_events: Union[Dict[str, List[Dict[str, Any]]], None],
         conf: Dict,
     ):
         super().__init__(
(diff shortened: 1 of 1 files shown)


turian:
"
In both the above cases (data with folds or not with folds), the grid point will first be found followed by using this best grid point configuration to train on the train+valid set. Thereafter, the final testing scores will be computed."

This should be optional, otherwise we are breaking the behavior that we used for the open tasks.

heareval/predictions/task_predictions.py
```diff
@@ -709,15 +714,40 @@ def dataloader_from_split_name(
...
-        metadata=metadata,
-    )
+    """
+    Get the dataloader for a `split_name` or a list of `split_name`
+
+    For a list of `split_name`, the dataset for each split will be concatenated.
+    The dataloader will be built from this concatenated
+    dataset
+    """
+    if isinstance(split_name, (list, set)):
```
turian, line 724: This is a bit weird, wouldn't it be better maybe to do this as two explicit functions?

heareval/predictions/task_predictions.py
```diff
@@ -749,30 +779,153 @@ def dataloader_from_split_name(
...
 
 
-def task_predictions_train(
+def task_predictions_train_test_full(
+    best_grid_point: GridPointResult,
     embedding_path: Path,
     embedding_size: int,
-    grid_points: int,
     metadata: Dict[str, Any],
+    data_splits: Dict[str, List[str]],
```
turian, line 802: This adds a lot of code but it's not clear what's the difference between this and below. That makes it quite confusing and I'm worried about the code duplication

heareval/predictions/task_predictions.py
```diff
@@ -1000,46 +1208,89 @@ def print_scores(grid_point_results: List[GridPointResult]):
...
-    test_results.update(
-        {
-            "validation_score": best_grid_point.validation_score,
-            "hparams": hparams_to_json(best_grid_point.hparams),
-            "postprocessing": best_grid_point.postprocessing,
-            "epoch": best_grid_point.epoch,
-            "time_in_min": best_grid_point.time_in_min,
-            "score_mode": best_grid_point.score_mode,
-            "embedding_path": str(embedding_path),
+    # Train the model using the best grid point configuration on the train+valid
```
turian, line 1218: This long function is getting a bit confusing, maybe we break it down into several functions? I am not sure tho. I am wondering if there is a way to do more code re-use to simplify things

turian:
We might want an alternate mode where, for each train/val/test set, we do a full grid search. (Not one grid search and keep the grid point for the other folds.) This might be useful for small datasets.

heareval/predictions/task_predictions.py
```diff
@@ -784,18 +937,20 @@ def task_predictions_train(
     start = time.time()
     predictor: AbstractPredictionModel
     if metadata["embedding_type"] == "event":
-        validation_target_events = json.load(
-            embedding_path.joinpath("valid.json").open()
-        )
-        test_target_events = json.load(embedding_path.joinpath("test.json").open())
+        # Only validation target event is required while searching for best grid point
```
jorshi, line 940: Are you sure? Don't we want to include early stopping on validation when we're training the final model as well?

heareval/predictions/task_predictions.py
jorshi: Agreed - I'm not really sure we need two different train functions. I understand the idea was to train the final model using the train + val to maximize the data we're using. But I think we still want to use the validation when training the final model and use early stopping.

heareval/predictions/task_predictions.py
jorshi: Yes. Can we instead treat everything as a folded dataset? And datasets that have predefined train, test, validation are just a special case where there is a single configuration? I think we could get rid of some of these conditionals that way.

jorshi (commented):
This is looking cool - I generally agree with Joseph's comments regarding code duplication. I think there is a close solution that is more simple.

heareval/predictions/task_predictions.py
```diff
@@ -749,30 +779,153 @@ def dataloader_from_split_name(
...
+    Model configurations are copied from the best grid point which was found using the
+    `train` and the `valid` split.
+    """
+    start = time.time()
+
+    # Get the configuration of the best_grid_point and use it to train on the
+    # full train+valid set and test on the test set
+    conf = best_grid_point.conf
+    # Set the max_epochs to the number of epochs done in the best grid point
+    conf["max_epochs"] = best_grid_point.epoch
```
jorshi, line 823: I see. Instead of using early stopping we cap the number of epochs to the number of epochs reached during the grid point search. Are we sure that's what we want? This is definitely quicker. But, I'm not positive this is what we want to do.

jorshi: @turian

turian: @jorshi @khumairraj No this is not correct.

You generally fix max_epochs when you are retraining and combining train+val so that you have a bigger train and no val any more.

If we are doing k-fold and fixing the hyperparameters, but we have separate train and validation folds, we should use our validation fold for early stopping again.

khumairraj:
Inherited by #315 . Closing this

PR #315, opened 2021-11-04
K-fold v2

jorshi:
Updates the hear eval pipeline to support K-Fold datasets. Depends on preprocessed data updates that are proposed in https://github.com/neuralaudio/hear-preprocess/pull/100 -- review those first. The small datasets that are used for tests in this will need to be updated in order for tests to run properly.

This builds off of @khumairraj's PR #310


heareval/predictions/task_predictions.py
```diff
@@ -997,49 +1121,77 @@ def print_scores(grid_point_results: List[GridPointResult]):
         grid_point_results.append(grid_point_result)
         print_scores(grid_point_results)
 
-    # Use the best model to compute test scores
+    # Use the best model to train remaining splits and compute test scores
```
jorshi, line 1124: This isn't totally accurate -- should be "Use the best hyperparameters to train models for remaining folds, then compute test scores using the resulting models"

heareval/predictions/task_predictions.py
```diff
@@ -700,7 +705,7 @@ def label_vocab_nlabels(embedding_path: Path) -> Tuple[pd.DataFrame, int]:
 
 
 def dataloader_from_split_name(
```
turian, line 707: Great :)

heareval/predictions/task_predictions.py
```diff
@@ -360,8 +360,8 @@ def __init__(
         nlabels: int,
         prediction_type: str,
         scores: List[ScoreFunction],
-        validation_target_events: Dict[str, List[Dict[str, Any]]],
-        test_target_events: Dict[str, List[Dict[str, Any]]],
+        validation_target_events: Union[Dict[str, List[Dict[str, Any]]], None],
```
turian, line 363: Wait why is the signature of this changing? I'm confused

turian: The addition of None here and below is confusing to me

jorshi: Ya I think this is a bug.

heareval/predictions/task_predictions.py
jorshi: All @khumairraj :)

heareval/predictions/task_predictions.py
```diff
@@ -904,6 +973,40 @@ def hparams_to_json(hparams):
     return {k: serialize_value(v) for k, v in hparams.items()}
 
 
+def data_splits_from_folds(folds: List[str]) -> List[Dict[str, List[str]]]:
```
turian, line 976: what's the string here? fold1, etc?

heareval/predictions/task_predictions.py
```diff
@@ -904,6 +973,40 @@ def hparams_to_json(hparams):
...
+    """
+    Create data splits by using Leave One Out Cross Validation strategy.
+
+    Test, Valid and Train Splits will consist of data folds. Successively treat
+    each fold as test split, the next fold as valid split and the remaining
+    fold as the train split.
+    Folds will be sorted before applying the above strategy
```
turian, line 983: to be documented here

heareval/predictions/task_predictions.py
```diff
@@ -904,6 +973,40 @@ def hparams_to_json(hparams):
...
+    Total data splits will be equal to n, n being the total number of folds.
+    Each fold will be tested by training on the remaining folds.
```
turian, line 985: I'd just write an explicit example of what is returned, key and values

turian: or say:

"""
With 5-fold, for example, we would have:
test=fold1, val=fold2, train=fold3..5,
test=fold2, val=fold3, train=fold4..5,1,
...

heareval/predictions/task_predictions.py
```diff
@@ -925,6 +1028,8 @@ def task_predictions(
         seed_everything(42, workers=False)
 
     metadata = json.load(embedding_path.joinpath("task_metadata.json").open())
+    metadata["mode"] = "folds"  # remove me, only for testing
```
turian, line 1031: What is the convention that was defined? Now I'm forgetting

heareval/predictions/task_predictions.py
```diff
@@ -997,49 +1123,77 @@ def print_scores(grid_point_results: List[GridPointResult]):
...
             "hparams": hparams_to_json(best_grid_point.hparams),
             "postprocessing": best_grid_point.postprocessing,
-            "epoch": best_grid_point.epoch,
-            "time_in_min": best_grid_point.time_in_min,
-            "score_mode": best_grid_point.score_mode,
+            "score_mode": split_grid_points[i].score_mode,
             "embedding_path": str(embedding_path),
         }
     )
+
```
turian, line 1195: The function is getting a bit long, can we break it down?

Can we save the scores of the different folds?

heareval/predictions/task_predictions.py
jorshi: Ya exactly. Will update the docstring

heareval/predictions/task_predictions.py
jorshi: Will update

heareval/predictions/task_predictions.py
jorshi: I don't think we have a convention defined yet - need to check this b/c metadata["mode"] might not actually be defined for the existing open tasks

heareval/predictions/task_predictions.py
jorshi: Did my best to pull some stuff out with out going crazy.

heareval/predictions/task_predictions.py
jorshi: Updated this so if "fold" is in metadata (and is a list of fold string names) then it will perform k-fold. Otherwise will work as normal with pre-defined splits.

khumairraj (approved):
Thanks for completing this @jorshi. It looks much cleaner. 
Also, there was some code duplication before, mostly around the behaviour of finding the best grid point and then retraining on all(train + Val) the data(without validation steps like early stopping, etc) with the characteristics of the best grid point(including the early stopping epoch and others). But, I see in this pr that we have removed that behavior. I think it was introducing lots of conditions, and it is good that we removed it.

heareval/predictions/runner.py
```diff
@@ -74,19 +74,16 @@ def runner(
...
 
-        embedding_size = json.load(open(train_embedding_dimensions))[1]
-        if (
-            embedding_size
-            != json.load(open(task_path.joinpath("valid.embedding-dimensions.json")))[1]
-            or embedding_size
-            != json.load(open(task_path.joinpath("test.embedding-dimensions.json")))[1]
-        ):
+        # Ensure all embedding sizes are the same across splits/folds
+        embedding_size = embedding_sizes[0]
```
turian, line 85: Do we even need `embedding_size` any more then? if we have `embedding_sizes`?

heareval/predictions/task_predictions.py
```diff
@@ -709,15 +709,40 @@ def dataloader_from_split_name(
...
-        embedding_type=embedding_type,
-        in_memory=in_memory,
-        metadata=metadata,
-    )
+    """
+    Get the dataloader for a `split_name` or a list of `split_name`
+
+    For a list of `split_name`, the dataset for each split will be concatenated.
+    The dataloader will be built from this concatenated
+    dataset
```
turian, line 717: explain the two cases with *examples*, so people understand why we need this complexity

heareval/predictions/task_predictions.py
```diff
@@ -766,13 +792,14 @@ def __init__(
...
         self.score_mode = score_mode
+        self.conf = conf
 
 
 def task_predictions_train(
     embedding_path: Path,
     embedding_size: int,
-    grid_points: int,
     metadata: Dict[str, Any],
+    data_splits: Dict[str, List[str]],
```
turian, line 802: What is data_splits? Maybe it's worth having a doc-string here? Or is this called by another function with the same set of parameters?

I am wondering if it even makes sense that we have a class for such a long list of things? Or most of them if it makes sense to divide it in a sensible way?

heareval/predictions/task_predictions.py
```diff
@@ -781,13 +808,26 @@ def task_predictions_train(
...
-        test_target_events = json.load(embedding_path.joinpath("test.json").open())
+
+        # Save the target events for validation and test
+        validation_target_events = {}
+        for split_name in data_splits["valid"]:
+            validation_target_events.update(
+                json.load(embedding_path.joinpath(f"{split_name}.json").open())
+            )
+        test_target_events = {}
+        for split_name in data_splits["test"]:
```
turian, line 826: Oh man this seems really confusing, particularly if you understand the other pattern from hear-preprocess that splits is a list, not a dict of lists. How do we fix this and make it clearer?

heareval/predictions/task_predictions.py
```diff
@@ -781,13 +808,26 @@ def task_predictions_train(
...
     if metadata["embedding_type"] == "event":
-        validation_target_events = json.load(
-            embedding_path.joinpath("valid.json").open()
-        )
-        test_target_events = json.load(embedding_path.joinpath("test.json").open())
+
+        # Save the target events for validation and test
+        validation_target_events = {}
+        for split_name in data_splits["valid"]:
+            validation_target_events.update(
```
turian, line 822: Why are we doing an update? how do we make sure we are not clobbering /. overwriting existing keys? This is so confusing

heareval/predictions/task_predictions.py
```diff
@@ -845,7 +885,7 @@ def task_predictions_train(
         logger=logger,
     )
     train_dataloader = dataloader_from_split_name(
-        "train",
+        data_splits["train"],
```
turian, line 888: Again this stuff is so confusing. Can we put all this weird logic in one place with a clear explanation, so this weird pattern doesn't occur throughout this file? The fewer patterns you have to remember, the better

heareval/predictions/task_predictions.py
```diff
@@ -904,6 +981,125 @@ def hparams_to_json(hparams):
     return {k: serialize_value(v) for k, v in hparams.items()}
 
 
+def data_splits_from_folds(folds: List[str]) -> List[Dict[str, List[str]]]:
+    """
+    Create data splits by using Leave One Out Cross Validation strategy.
+
+    folds is a list of dataset partitions created during pre-processing. For example,
+    for 5-fold cross val: ["fold00", "fold01", ..., "fold04"]. This function will create
```
turian, line 989: Is this the final naming convention that we settled upon?

heareval/predictions/task_predictions.py
```diff
@@ -904,6 +981,125 @@ def hparams_to_json(hparams):
...
+
+def sort_grid_points(grid_point_results: List[GridPointResult]) -> None:
+    """
+    Sort grid point results in place, so that the first result
+    is the best.
+    """
+    # TODO: Assert all score modes are the same?
+    mode = grid_point_results[0].score_mode
+    # Pick the model with the best validation score
+    grid_point_results.sort(key=lambda g: -g.validation_score)
```
turian, line 1078: This is kinda gross and confusing.

Rewrite as:

```
    if mode == "max":
        grid_point_results.sort(key=lambda g: g.validation_score)
    elif mode == "min":
        grid_point_results.sort(key=lambda g: g.validation_score, reversed=True)
```

heareval/predictions/task_predictions.py
```diff
@@ -904,6 +981,125 @@ def hparams_to_json(hparams):
...
+        data_splits = data_splits_from_folds(folds)
+    else:
+        raise AssertionError(
+            f"Unknown split_mode: {metadata['split_mode']} in task metadata."
+        )
+
+    return data_splits
+
+
+def sort_grid_points(grid_point_results: List[GridPointResult]) -> None:
```
turian, line 1070: Why are we sorting in-place? Why not return the sorted copy, and overwrite it when in the return? Nicer not to have side effects.

heareval/predictions/task_predictions.py
```diff
@@ -904,6 +981,125 @@ def hparams_to_json(hparams):
...
+        grid_point_results.reverse()
+    else:
+        raise ValueError(f"mode = {mode}")
+
+
+def print_scores(grid_point_results: List[GridPointResult], embedding_path: Path):
+    sort_grid_points(grid_point_results)
+    for g in grid_point_results:
+        print(
+            json.dumps(
```
turian, line 1091: Why not make a GridPointResult.str or GridPointResult.json() method, to simplify?

heareval/predictions/task_predictions.py
khumairraj: yes

heareval/predictions/task_predictions.py
turian: I'm still kinda concerned tho, it's really long. Any way we can go further?

heareval/predictions/task_predictions.py
khumairraj: We are doing an update since the target event might now be coming from a set of splits

heareval/predictions/task_predictions.py
khumairraj: #319

heareval/predictions/task_predictions.py
khumairraj: # 319

heareval/predictions/task_predictions.py
khumairraj: #319

heareval/predictions/task_predictions.py
khumairraj: #320

turian (approved):

PR #325, opened 2021-11-14
Variable Audio Embedding Generation | ADD LWLRAP, mAP, D_prime

khumairraj:


heareval/score.py
```diff
@@ -223,6 +225,83 @@ class EventBasedScore(SoundEventScore):
     score_class = sed_eval.sound_event.EventBasedMetrics
 
 
+class MeanAveragePrecision(ScoreFunction):
```
turian, line 228: If you're going to use someone else's code, you should reference them and also indicate where the code is from

heareval/score.py
```diff
@@ -223,6 +225,100 @@ class EventBasedScore(SoundEventScore):
...
+        per_class_lwlrap = np.sum(
+            precisions_for_samples_by_classes, axis=0
+        ) / np.maximum(1, labels_per_class)
+        return per_class_lwlrap
+
+    def __call__(self, predictions: np.ndarray, targets: np.ndarray, **kwargs) -> float:
+        assert predictions.ndim == 2
+        assert targets.ndim == 2  # One hot
+        per_class_lwlrap = self.calculate_per_class_lwlrap(targets, predictions)
+        return np.mean(per_class_lwlrap)
```
jorshi, line 319: How does this differ than the google colab? Are we macro averaging as opposed to the weighted averaging? Is that what Eduardo suggested for lwlwrap? Maybe make a quick comment above about what has changed? Also, do we have a reference for lwlwrap any where -- it's hard to tell what this is doing by looking at the code, a brief description and reference would be awesome

heareval/score.py
jorshi: I think this is resolved now?

turian (approved):

PR #327, opened 2021-11-16
Prediction output

jorshi:
Optional to specify the output directory for the predictions.

### DIFF
--- a/heareval/predictions/runner.py
+++ b/heareval/predictions/runner.py
@@ -90,7 +90,7 @@ def runner(
             # A separate output directory was passed in.
             # Create a subdirectory within that with the same name as
             # the task to save the results in.
-            outdir = Path(out_dir).joinpath(task_path.name)
+            outdir = Path(outdir).joinpath(task_path.name)
             outdir.mkdir(parents=True, exist_ok=True)
 
         # Get embedding sizes for all splits/folds
--- a/heareval/predictions/runner.py
+++ b/heareval/predictions/runner.py
@@ -85,14 +85,17 @@ def runner(
         # Get the output directory name
         output_path: Path
         if outdir is None:
-            # Default is the task_path -- dir with embeddings
-            output_path = task_path
+            # Default is a folder predictions/embedding_name/task_name
+            output_path = Path("predictions").joinpath(task_path.parent.name)
+            output_path = output_path.joinpath(task_path.name)
         else:
             # A separate output directory was passed in.
             # Create a subdirectory within that with the same name as
             # the task to save the results in.
             output_path = Path(outdir).joinpath(task_path.name)
-            output_path.mkdir(parents=True, exist_ok=True)
+
+        # Create the output path if it needs to be created
+        output_path.mkdir(parents=True, exist_ok=True)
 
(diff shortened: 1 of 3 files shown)


heareval/score.py
```diff
@@ -190,7 +190,7 @@ def sed_eval_event_container(
                 reference_events.append(
                     {
                         # Convert from ms to seconds for sed_eval
-                        "event_label": event["label"],
+                        "event_label": str(event["label"]),
```
jorshi, line 193: sed_eval requires event labels to be strings -- this was breaking on maestro

heareval/predictions/runner.py
```diff
@@ -23,6 +23,13 @@
     nargs=-1,
     required=True,
 )
+@click.option(
+    "--outdir",
+    default=None,
+    help="Output directory to save prediction results. (Defaults to "
```
turian, line 29: Seems like it will (and SHOULD) be a separate directory from input embeddings:

```
predictions/embedding_name/task_name
```

heareval/predictions/runner.py
```diff
@@ -23,6 +23,13 @@
     nargs=-1,
     required=True,
 )
+@click.option(
+    "--outdir",
```
turian, line 27: Can we call this --predictions-dir ? Or similar naming convention as embed which uses --embeddings-dir as the output dir?

heareval/predictions/runner.py
```diff
@@ -74,6 +82,21 @@ def runner(
         if not task_path.is_dir():
             raise ValueError(f"{task_path} should be a directory")
 
+        # Get the output directory name
```
turian, line 85: Uhm.... you should be branching off `leader` branch. Otherwise there is no way to re-use my code that checks if prediction has already been done and we can skip it for this model + prediction

heareval/predictions/runner.py
```diff
@@ -74,6 +82,21 @@ def runner(
...
+        output_path: Path
+        if outdir is None:
+            # Default is a folder predictions/embedding_name/task_name
+            output_path = Path("predictions").joinpath(task_path.parent.name)
+            output_path = output_path.joinpath(task_path.name)
+        else:
+            # A separate output directory was passed in.
+            # Create a subdirectory within that with the same name as
+            # the task to save the results in.
+            output_path = Path(outdir).joinpath(task_path.name)
```
turian, line 95: In embeddings, if the embedding dir exists but is incomplete, we remove it, just so as not to have mixed results. We should do that here too. (Unless it's the embedding dir. TBH I think embedding dir should never equal predictions dir.)

heareval/predictions/runner.py
jorshi: This was an old commented -- updated

heareval/predictions/runner.py
jorshi: Yup. My bad. Updated.

heareval/predictions/runner.py
jorshi: Updated to remove the predictions dir if it is incomplete. Also added assertion to verify it is not the same as the input.

PR #331, opened 2021-11-20
Add event based metrics

khumairraj:


heareval/predictions/task_predictions.py
```diff
@@ -865,6 +874,18 @@ def _combine_target_events(split_names: List[str]):
...
+        # best task specific postprocessing, can be task
+        # specific, present in the task metadata in
+        # evaluation_params.postprocessing_grid. If not, the default
+        # EVENT_POSTPROCESSING_GRID will be used.
+        if "event_postprocessing_grid" in metadata["evaluation_params"]:
+            postprocessing_grid = metadata["evaluation_params"][
+                "event_postprocessing_grid"
+            ]
+        else:
+            postprocessing_grid = EVENT_POSTPROCESSING_GRID
```
turian, line 887: EVENT_POSTPROCESSING_GRID can we just remove this and have a specific dcase grid and no default? If this is tricky let's just make an issue for it.

heareval/score.py
```diff
@@ -239,6 +239,28 @@ class EventBasedScore(SoundEventScore):
...
     ),
+    "event_onset_50ms_fms": partial(
+        EventBasedScore,
+        name="event_onset_50ms_fms",
+        score="f_measure",
+        params={
+            "evaluate_onset": True,
+            "evaluate_offset": False,
+            "t_collar": 0.05,
+            "percentage_of_length": 0.5,
```
turian, line 250: Remove this. percentage_of_length is not used when evaluate_offset is false.

khumairraj: Should we also remove from the existing 200ms fms?

heareval/score.py
```diff
@@ -239,6 +239,28 @@ class EventBasedScore(SoundEventScore):
...
+        },
+    ),
+    "event_onset_offset_50ms_20perc_fms": partial(
+        EventBasedScore,
+        name="event_onset_offset_50ms_20perc_fms",
```
turian, line 255: In MAESTRO paper (and our doc), they use fms + prc + rcl. Is it possible (and not too hard) to compute all three without running sed_eval three times (which is slow)? Maybe we could even run EventBasedScore just once for both onset f+p+r and onset+offset f+p+r. Like one sed_eval for all six?

heareval/predictions/task_predictions.py
khumairraj: #332

heareval/score.py
khumairraj: #333

PR #334, opened 2021-11-21
Allow metrics to return tuple of scores

khumairraj:


heareval/score.py
```diff
@@ -142,24 +170,25 @@ class SoundEventScore(ScoreFunction):
...
         if params is None:
             params = {}
         super().__init__(label_to_idx=label_to_idx, name=name, maximize=maximize)
-        self.score = score
+        self.scores = scores
         self.params = params
         assert self.score_class is not None
 
-    def __call__(self, predictions: Dict, targets: Dict, **kwargs):
+    def compute(self, predictions: Dict, targets: Dict, **kwargs):
```
turian, line 191: Should we add `-> float` to the return type here?

heareval/score.py
```diff
@@ -53,6 +54,28 @@ def label_to_binary_vector(label: List, num_labels: int) -> torch.Tensor:
     return binary_labels
```
turian, line 56: So I've seen us do this a few times, and I am not sure I love this pattern. That we have simple types and then, changing to more complex types, we *overload* the existing function and then use `isinstance` for runtime typing. It seems more fragile than breaking a function into something with a very specific return type, which we can statically check using `mypy`.

Do you think we can fix this easily? If not, let's make an issue to revisit this?

heareval/score.py
```diff
@@ -53,6 +54,28 @@ def label_to_binary_vector(label: List, num_labels: int) -> torch.Tensor:
...
 
 
+def validate_score_return_type(
+    ret: Union[float, int, Tuple[Tuple[str, Union[int, float]]]]
```
turian, line 58: When would we ever return an int? The return type for a score should always be float

heareval/score.py
turian: I guess this is okay, since it's only two types if we remove the ints.

heareval/score.py
```diff
@@ -76,7 +99,12 @@ def __init__(
             self.name = name
         self.maximize = maximize
 
-    def __call__(self, predictions: Any, targets: Any, **kwargs) -> float:
+    def __call__(self, *args, **kwargs) -> float:
```
turian, line 102: Wait so why is `return -> float` if in some cases it can be `Tuple[Tuple[str, float]]`? Where do `Tuple[str, float]` come from? (And if not's here, can you document in `validate_score_return_type` where the tuples might come from?

heareval/score.py
```diff
@@ -76,7 +99,12 @@ def __init__(
...
+        ret = self.compute(*args, **kwargs)
+        validate_score_return_type(ret)
```
turian, line 104: feels like the type signature is probably wrong here but I am not sure?
Have you run mypy by hand on heareval in this branch?

heareval/score.py
turian: Also please add a comment explaining why we have __call__ and compute, and why

heareval/score.py
```diff
@@ -176,10 +205,18 @@ def __call__(self, predictions: Dict, targets: Dict, **kwargs):
...
-        return overall_scores[self.score][self.score]
+        # results_overall_metrics return a pretty large nested selection of scores,
+        # with dicts of scores keyed on the type of scores, like f_measure, error_rate,
+        # accuracy
+        nested_overall_scores: Dict[
+            str, Dict[str, float]
+        ] = scores.results_overall_metrics()
+        # Open up nested overall scores
+        overall_scores: Dict[str, Union[float, int]] = dict(
+            ChainMap(*nested_overall_scores.values())
```
turian, line 216: cool did not know chain map

heareval/score.py
```diff
@@ -176,10 +205,18 @@ def __call__(self, predictions: Dict, targets: Dict, **kwargs):
...
-        # Keep the specific score we want
-        return overall_scores[self.score][self.score]
+        # results_overall_metrics return a pretty large nested selection of scores,
+        # with dicts of scores keyed on the type of scores, like f_measure, error_rate,
+        # accuracy
+        nested_overall_scores: Dict[
+            str, Dict[str, float]
+        ] = scores.results_overall_metrics()
+        # Open up nested overall scores
+        overall_scores: Dict[str, Union[float, int]] = dict(
```
turian, line 215: Are you sure we want dict? Can't we keep the scores in the exact order of `scores` param? Seems safer that way.

heareval/predictions/task_predictions.py
```diff
@@ -253,6 +254,40 @@ def validation_step(self, batch, batch_idx):
...
+        assert hasattr(self, "scores"), "Scores for the model should be defined"
+        end_scores = {}
+        for score in self.scores:
+            score_ret = score(*score_args)
+            validate_score_return_type(score_ret)
+            # If the returned score is a tuple, store each subscore as separate entry
+            if isinstance(score_ret, tuple):
+                # The first score in the returned tuple will be used
+                # as primary score for this metric
+                end_scores[f"{name}_{score}"] = score_ret[0][1]
```
turian, line 268: Not sure if this is the right PR to change this, but I hate this JSON schema I defined.

It looks like this:

```
    "test": {
        "test_score": 0.5070093274116516,
        "test_loss": 11.01280689239502,
        "test_event_onset_200ms_fms": 0.5070093274116516,
        "test_segment_1s_er": 0.5021833777427673,
        "validation_score": 0.4511459767818451,
        "epoch": 470,
        "time_in_min": 64.55239222049713
    },
```

When I want to crawl over this JSON and find ALL the scores, I resort to the hacky trick of seeing if the string starts with "test" and isn't "test_score" or "test_loss". Kinda gross.

I would prefer ["test"]["test_scores"] = {"onset_fms": ..., ...} etc in addition to ["test"]["test_score"] and ["test"]["test_loss"].

Maybe we should make an issue for this?

heareval/predictions/task_predictions.py
```diff
@@ -253,6 +254,40 @@ def validation_step(self, batch, batch_idx):
...
+        """Logs the metric score value for each score defined for the model"""
+        assert hasattr(self, "scores"), "Scores for the model should be defined"
+        end_scores = {}
+        for score in self.scores:
+            score_ret = score(*score_args)
+            validate_score_return_type(score_ret)
+            # If the returned score is a tuple, store each subscore as separate entry
+            if isinstance(score_ret, tuple):
+                # The first score in the returned tuple will be used
+                # as primary score for this metric
```
turian, line 267: Wait. No. WTF.

We don't have a primary score per metric. We have a primary score AKA optimization criterion AKA target score for each *task*.

heareval/predictions/task_predictions.py
```diff
@@ -253,6 +254,40 @@ def validation_step(self, batch, batch_idx):
...
+            else:
+                raise ValueError(
+                    f"Return type {type(score_ret)} is unexpected. Return type of "
+                    "the score function should either be a "
+                    "tuple(tuple(str, (float, int))) or float or int. "
+                )
+
+        # Weird, scores can be nan if precision has zero guesses
+        for score_name, value in end_scores.items():
+            end_scores[score_name] = 0.0 if math.isnan(value) else value
```
turian, line 283: Wait. No. Why would we do this? NaN is NaN. It's wrong to convert it to zero. Particularly for a score we want to minimize not maximize. This is bad.

heareval/predictions/task_predictions.py
khumairraj: by priamry score at the metric level, i meant the fms will be primary in any of the optimisation or will behave as the target

heareval/score.py
khumairraj: we get the dict. The final return is in the order of the scores param

heareval/score.py
khumairraj: return type is not float but a Tuple[Tuple[str, float]]. added

heareval/predictions/task_predictions.py
khumairraj: #336

heareval/predictions/task_predictions.py
khumairraj: added information for this in `validate_score_return_type`

heareval/predictions/task_predictions.py
khumairraj: removing this ( was there from before )

heareval/score.py
khumairraj: fixed

heareval/predictions/task_predictions.py
```diff
@@ -253,6 +253,38 @@ def validation_step(self, batch, batch_idx):
...
+        end_scores = {}
+        for score in self.scores:
+            score_ret = score(*score_args)
+            validate_score_return_type(score_ret)
+            # If the returned score is a tuple, store each subscore as separate entry
+            if isinstance(score_ret, tuple):
+                # The first score in the returned tuple will be used for
+                # any optimisation criterion, for instance if the metric is
+                # primary metric, the score corresponding to the first tuple
+                # will be used for early stopping
```
turian, line 268: This still isn't right. The first score in the returned tuple for the first self.scores is the optimization criterion.

The comment suggests that the first score in the returned tuple for EVERY self.scores is the optimization criterion

heareval/score.py
```diff
@@ -76,9 +111,23 @@ def __init__(
             self.name = name
         self.maximize = maximize
 
-    def __call__(self, predictions: Any, targets: Any, **kwargs) -> float:
+    def __call__(self, *args, **kwargs) -> Union[Tuple[Tuple[str, float], ...], float]:
```
turian, line 114: https://github.com/neuralaudio/hear-eval-kit/issues/339

turian (approved):

jorshi (approved):

heareval/predictions/task_predictions.py
```diff
@@ -254,6 +254,35 @@ def validation_step(self, batch, batch_idx):
     def test_step(self, batch, batch_idx):
         return self._step(batch, batch_idx)
 
+    def log_scores(self, name: str, score_args):
+        """Logs the metric score value for each score defined for the model"""
+        assert hasattr(self, "scores"), "Scores for the model should be defined"
+        end_scores = {}
+        # The first score in the first `self.scores` is the optimization criterion
+        for score in self.scores:
```
turian, line 262: We probably only need to do this for the *primary* score, otherwise we are wasting time during validation.

https://github.com/neuralaudio/hear-eval-kit/issues/344

PR #335, opened 2021-11-21
Save Predictions + Add prediction logging

khumairraj:


heareval/predictions/runner.py
```diff
@@ -18,6 +19,24 @@
 from heareval.predictions.task_predictions import task_predictions
 
 
+def get_logger(task_name, log_path):
```
turian, line 22: Can you type the return here, and everywhere that logger is passed around?

heareval/predictions/runner.py
```diff
@@ -107,7 +129,7 @@ def runner(
         )
```
turian, line 129: Why aren't we passing in the logger here?

heareval/predictions/task_predictions.py
```diff
@@ -1208,6 +1213,7 @@ def task_predictions(
         seed_everything(42, workers=False)
 
     metadata = json.load(embedding_path.joinpath("task_metadata.json").open())
+    logger = logging.getLogger(metadata["task_name"])
```
turian, line 1216: Instead of creating this again (which means it is created in twice places and thus the chance of bugs increase), why not just pass it in here?

I think my preference is that `logging.getLogger` is called only once per task, in runner.py, and everywhere else we pass it around the functions (with the parameter typed)

heareval/predictions/runner.py
khumairraj: we get the logger with the help of the task name

heareval/predictions/task_predictions.py
```diff
@@ -1319,6 +1326,20 @@ def task_predictions(
...
+                split_grid_points[i].predictor.test_predicted_events,
+                embedding_path.joinpath(f"{test_fold_str}.predictions.json").open("w"),
+                indent=4,
+            )
+        pickle.dump(
+            split_grid_points[i].predictor.test_predicted_labels,
+            open(
+                embedding_path.joinpath(f"{test_fold_str}.predicted-labels.pkl"), "wb"
+            ),
+        )
```
turian, line 1341: a) We can't do any interesting downstream analysis unless predictions AND targets are pickled. This MUST be fixed.

b) Is it difficult to pickle these scores where the test stuff is actually computed, rather than relying upon nebulous global state that may change? (For b) maybe make an issue and move on.)

turian: By global state, I mean split_grid_points[i].predictor.test_predicted_labels. Maybe this is okay. but we still must fix a)

khumairraj: Targets were already saved from before by the embedding module. @jorshi Thanks for adding it here as well. Will be in one place.

khumairraj: For b) #340

turian (approved):

PR #363, opened 2022-04-28
Update task_predictions.py

ltetrel:
Fixed parameters in trainer.test for pytorch >1.6


jorshi:
Failing DCASE test is fixed in #364

